PR #7, opened 2023-12-12
Add IRewardModifyingZone

modargo:
This is the first cut of `IRewardModifyingZone`. I put this together pretty quickly so I haven't tested all the small interactions and edge cases yet, but the core should work.

I built this with a few thoughts in mind:
* Provide a simple way of doing common operations
* Start with a core set on methods with high flexibility so that a broad range of use cases are supported with a bit more work
* Make it hard to screw things up in a way that makes saving and reloading change the rewards

The documentation comments plus the example will hopefully be enough to understand what's going on. 

For the overall shape of the API, I took a look through the ideas spreadsheet and this should meet the needs of many of the reward modifiers listed so far. There are some potential extensions to this that I considered but didn't put in the initial version:
* Methods that hook into the normal relic and potion drop logic to let zones define their own distribution. There are ways to achieve this with the current hooks, by replacing the generated relic/potion with your own, but it might be nice to have more direct support. A moderate number of zones would benefit from this.
* A method for not being able to skip card rewards. This may be needed for Dayvig's Scrapyard zone ("You gain 2 cards rewards from combat, but cannot skip card rewards (anywhere)").
* Any support specific to `CustomReward`. They can be added in the methods for getting additional rewards, and I didn't see a clear need for anything beyond that. TBD how much people even want to make custom rewards, the only things that jumped out at me as maybe needing them are from Rucodeby's Laser Area ("Chance to take a modifier for cards that enchance their damage instead card reward") and Vex's Hatlands ("New reward type: Hat.").

None of these are obviously needed to me, so I thought it made sense to discuss before adding any of them. Once the foundation added by this PR is in place, it won't be hard to add new methods, and we can work out what we want to have before opening things up for contributions.

### DIFF
--- a/src/main/java/spireMapOverhaul/patches/interfacePatches/RewardModifierPatches.java
+++ b/src/main/java/spireMapOverhaul/patches/interfacePatches/RewardModifierPatches.java
@@ -1,5 +1,6 @@
 package spireMapOverhaul.patches.interfacePatches;
 
+import basemod.ReflectionHacks;
 import com.evacipated.cardcrawl.modthespire.lib.*;
 import com.megacrit.cardcrawl.cards.AbstractCard;
 import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
@@ -7,6 +8,8 @@ import com.megacrit.cardcrawl.rewards.RewardItem;
 import com.megacrit.cardcrawl.rooms.AbstractRoom;
 import com.megacrit.cardcrawl.saveAndContinue.SaveAndContinue;
 import com.megacrit.cardcrawl.screens.CombatRewardScreen;
+import com.megacrit.cardcrawl.vfx.FastCardObtainEffect;
+import com.megacrit.cardcrawl.vfx.cardManip.ShowCardAndObtainEffect;
 import javassist.CtBehavior;
 import spireMapOverhaul.abstracts.AbstractZone;
 import spireMapOverhaul.util.Wiz;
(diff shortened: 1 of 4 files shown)


src/main/java/spireMapOverhaul/patches/interfacePatches/RewardModifierPatches.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @SpirePatch2(clz = AbstractRoom.class, method = "update")
+    public static class ModifyRewardsPatch {
+        @SpireInsertPatch(locator = Locator.class)
+        public static void ModifyRewards(AbstractRoom __instance) {
+            AbstractZone zone = Wiz.getCurZone();
+            if (zone instanceof IRewardModifyingZone) {
+                // We don't want to affect uses of CombatRewardScreen that are in non-battle events such as Sensory Stone
+                if (AbstractDungeon.getCurrMapNode() != null && AbstractDungeon.getCurrMapNode().room != null && AbstractDungeon.getCurrMapNode().room.isBattleOver) {
```
modargo, line 59: This design choice is worth noting. The reward modifiers in the spreadsheet are generally associated with combats, but there are some zones where it's ambiguous. E.g. for Tetraminus's Broken Space or Darkglade's Heavenly Clouds, what happens if you get Sensory Stone while in the zone, or purchase Orrery? Do those rewards get the modifier applied?

The answer being "no" is most intuitive to me. However, it wouldn't be too had to change if we wanted to allow this. We'd add a method like `affectNonCombatRewards` and check that here.

src/main/java/spireMapOverhaul/patches/interfacePatches/RewardModifierPatches.java
```diff
@@ -0,0 +1,127 @@
...
+                    AbstractDungeon.combatRewardScreen.rewards.addAll(z.getAdditionalRewards());
+                    z.modifyRewards(AbstractDungeon.combatRewardScreen.rewards);
+                }
+            }
+        }
+
+        private static class Locator extends SpireInsertLocator {
+            @Override
+            public int[] Locate(CtBehavior ctMethodToPatch) throws Exception {
+                // We ensure that this happens after the save file is made, outside of any checks for loading saves
```
modargo, line 76: This is how we avoid save/reload inconsistency.

It's still possible to create save/reload inconsistency by using other RNG's in the card methods, because the weird thing about card rewards is that when you win a fight, the card rewards are created before the save file... but when you save and load, the card rewards are created again, based on the save file. 

The base game is able to do this without causing issues because it has special code to roll back the card RNG; see use of `AbstractDungeon.cardRng.counter` in `AbstractRoom.update` if you're curious.

The typical way things go wrong is when mods patch into the card creation logic and add other RNG calls. Effectively, that makes saving and loading advance RNG counters a second time. Depending on exactly what a mod is doing, there are other ways to get this right, but also other ways to get this wrong. It's a tricky area of code.

src/main/java/spireMapOverhaul/zoneInterfaces/IRewardModifyingZone.java
```diff
@@ -0,0 +1,84 @@
...
+     * Note that the list of rewards will include any additional rewards you added and any modifications you made.
+     * @param rewards The set of rewards.
+     */
+    default void modifyRewards(ArrayList<RewardItem> rewards) {}
+
+    /**
+     * Hook for triggering effects when a card is added to the deck.
+     * @param card The card added.
+     */
+    default void onObtainCard(AbstractCard card) {}
```
modargo, line 74: This method and `changeRareCardRewardChance` weren't part of my initial plan, but they looked like they will be needed for a few zones, and they both have base game equivalents that are relic hooks, so I went ahead and added them. 

The specific zones that motivated this are:
* Cany's Shocked Waters zone ("You are twice as likely to find Rare cards, but picking up a Rare card will cause you to lose 3 HP.")
* Tetraminus's Void Seed zone ("Rewards Cost HP but are Higher rarity.")
* Blank's Infinite Void zone ("High chance to drop a rare card reward instead of a normal card reward.")

PR #12, opened 2023-12-19
EncounterModifyingZone interface, example zone using it, and Bestiary integration

modargo:
This adds `EncounterModifyingZone` to support use cases including:
* Defining your own normal and/or elite enemy list for the zone
* Adding additional monsters to encounters in the zone
* Changing encounters in more complex ways that require you to implement substantial logic yourself

My thought with this as to build something that meets the needs of the majority of straightforward use cases, aka items 1 and 2 above, with a particular focus on the first use case given how many zones want to do it. This should simplify the work most people need to do. In particular, the way this is set up means that zone creators won't need to write `BaseMod.addEncounter` calls, logic to avoid duplication, or encounter weights. On the other hand, if you want something more complex than "here's the list of monsters in my zone by act", your use case isn't entirely supported. Based on the zones ideas listed I think that's a reasonable trade-off.

Alongside that, this PR contains a few more additions;
* An example zone, which is fairly extensive. It includes a number of custom monsters, which are all adapted from my act mods.
* Bestiary integration, which is straightforward. The patch is adapted from the original that I wrote for my act mods.

I'm going to write comments for a few things that are worth calling out.


src/main/java/spireMapOverhaul/patches/interfacePatches/EncounterModifierPatches.java
```diff
@@ -0,0 +1,145 @@
...
+        @SpirePostfixPatch
+        public static MonsterGroup ChangeEncounter(AbstractDungeon __instance, MonsterGroup __result) {
+            MonsterGroup result = __result;
+            AbstractZone zone = Wiz.getCurZone();
+            if (zone instanceof EncounterModifyingZone) {
+                EncounterModifyingZone z = (EncounterModifyingZone)zone;
+                List<EncounterModifyingZone.ZoneEncounter> encounters = z.getNormalEncounters();
+                if (encounters != null) {
+                    encounters = encounters.stream().filter(e -> e.getActNum() == ActUtil.getRealActNum()).collect(Collectors.toList());
+                    if (LastZoneNormalEncounter.lastZoneNormalEncounter != null && LastZoneNormalEncounter.lastZoneNormalEncounter.startsWith(zone.id + ":")) {
```
modargo, line 34: I've built this in under the assumption that all the zone that define their own encounter list will want the standard no-duplicate-encounters behavior. It's something people don't always realize exist, but I think of it as a core part of how encounters work under the hood in Slay the Spire.

If some zone wants to bypass this, it's easy enough for us to add a method to turn it off, but I don't see a reason to encourage it.

src/main/java/spireMapOverhaul/zoneInterfaces/EncounterModifyingZone.java
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Defines a new encounter in a zone.
+     * An encounter requires an ID, which should be the same as the monster ID for single enemy encounters. This is
+     * needed by BaseMod (and can be used to fight your new encounter from the console with the fight command).
+     * An encounter also requires an act number, and will only show up in that act.
+     * The enemies in an encounter are defined by a function that returns either a single monster or a MonsterGroup.
+     * An encounter may have a special name. If no name is provided, the encounter will be named based on the enemies.
+     */
+    public static class ZoneEncounter {
```
modargo, line 51: This class is how an encounter is defined.

The base game uses weights so that different encounters show up with different probabilities. That didn't seem necessary for any of the zones in the document, so I left it out. It would be easy to add in as part of the this class.

src/main/java/spireMapOverhaul/zones/example/monsters/DreadMoth.java
```diff
@@ -0,0 +1,105 @@
+package spireMapOverhaul.zones.example.monsters;
```
modargo, line 1: I recall us discussing that rather than have separate top level packages for powers/cards/monsters/etc., each zone should just have those classes in their zone package. That's what I've done here.

For all the monsters I've added, since they're adapted from my existing act mods, I did not translate them to use the various methods in `AbstractSMOMonster` that are intended to make setting up moves easier.

src/main/java/spireMapOverhaul/zones/example/ExampleEncounterModifyingZone.java
```diff
@@ -0,0 +1,112 @@
...
+        return Color.BLUE.cpy();
+    }
+
+    @Override
+    public boolean canSpawn() {
+        return this.isAct(1) || this.isAct(2);
+    }
+
+    @Override
+    public List<ZoneEncounter> getNormalEncounters() {
```
modargo, line 55: I intend to use all the act 1 stuff here for my Invasion zone.

src/main/java/spireMapOverhaul/patches/interfacePatches/EncounterModifierPatches.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static void addAdditionalMonsters(EncounterModifyingZone zone, MonsterGroup monsterGroup) {
+        List<AbstractMonster> additionalMonsters = zone.getAdditionalMonsters();
+        if (additionalMonsters != null) {
+            float currentX = monsterGroup.monsters.stream().map(m -> m.drawX).min(Comparator.comparingDouble(x -> (double)x)).orElse(0.0f);
+            for (AbstractMonster monster : additionalMonsters) {
+                // We give healthy additional space to account for fights such as Reptomancer, which spawn minions in
+                // the empty space near the main monster. Since there are modded enemies with this behavior too, this
+                // isn't perfect, but it's pretty much impossible to get perfect positioning for this given all the
+                // modded enemies out there and other ways that enemies can be added to fights.
```
modargo, line 97: I've seen various attempts to do this (positioning new enemies dynamically). If you know of one that you think is more robust than this, we can switch over to it.

src/main/java/spireMapOverhaul/util/ActUtil.java
```diff
@@ -0,0 +1,11 @@
+package spireMapOverhaul.util;
+
+import com.megacrit.cardcrawl.core.Settings;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+
+public class ActUtil {
+    public static int getRealActNum() {
```
modargo, line 7: I extracted this out of AbstractZone because I wanted it to stay consistent if we had to modify this later.

For example, I'm worried that there's a bug with this if you're in endless mode and decide to go to act 4+. Unless `Settings.endless` gets set to false when you enter act 4, won't this say you're in act 1 again? Didn't have time to test this, but having the logic in one place means we only need to fix one place if there are any bugs like that.

src/main/java/spireMapOverhaul/SpireAnniversary6Mod.java
```diff
@@ -217,6 +221,14 @@ public void receivePostInitialize() {
         TextCodeInterpreter.addAccessible(ZoneShapeMaker.class);
     }
 
+    public static void addMonsters() {
+        for (AbstractZone zone : allZones) {
```
erasels, line 225: I think we should work with unfilteredAllZones here, right? Seems rpoblematic not to load encounters from zones that are currently deactiveated in case they get activated again

src/main/java/spireMapOverhaul/zoneInterfaces/EncounterModifyingZone.java
```diff
@@ -0,0 +1,122 @@
...
+
+public interface EncounterModifyingZone {
+    /**
+     * Defines the set of normal encounters for the zone, replacing all normal combats.
+     * If this is non-null, normal encounters in the zone will be randomly selected from this list.
+     * No normal encounter will be fought twice in a row; if only one is defined, the zone will fall back to the
+     * encounters defined by the current act.
+     * @return The list of normal encounters for the zone.
+     */
+    default List<ZoneEncounter> getNormalEncounters() { return null; }
```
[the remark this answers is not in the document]
erasels: I think just making sure that zone encounters (if they add new ones) are overtuned aka not harder than hard but not easier than easy, we should be fine. Balance will likely be a bit wacky anyway, don#t see a reasont to limit generation this strictly for what will likely be a small gain. However if it turns out new encouters are problematic when you see them early, i'd be amendable to that change.

src/main/java/spireMapOverhaul/zoneInterfaces/EncounterModifyingZone.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    /**
+     * Lets you change the encounter however you want.
+     * You should not need to use this unless you're doing something unusual.
+     * @param monsterGroup The encounter.
+     * @return The changed encounter.
+     */
+    default MonsterGroup changeEncounter(MonsterGroup monsterGroup) { return monsterGroup; }
```
erasels, line 99: Do you have an example for when you'd use this?

src/main/java/spireMapOverhaul/SpireAnniversary6Mod.java
modargo: Yeah, I just overlooked that we'd merged in the filtering.

src/main/java/spireMapOverhaul/zoneInterfaces/EncounterModifyingZone.java
modargo: All the examples I thought of are somewhat contrived:
* You want to do your own enemy positioning
* You want to replace a specific enemy with a different enemy
* You want to implement something that does your own tracking of encounters to decide what to replace each encounter with

This was the last thing I added, and I saw it as an fallback in case some zone needed to do something that wasn't anticipated by the rest of the API.

src/main/java/spireMapOverhaul/zoneInterfaces/EncounterModifyingZone.java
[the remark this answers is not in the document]
modargo: We'll see how it goes!

modargo:
Added two other fixes to this:
* I think tooltips were added after I pulled code to start work on this, so I needed to fix that for my new zone
* Changed two uses of allZones (one in my code, one in the keyword loading code) that I noticed to use unfilteredAllZones instead

PR #14, opened 2023-12-20
Show the zone for each node in the run history screen

modargo:
I like run history having information like this, so I went ahead and implemented it. The patch is adapted from the patches I've written for stuff like this in other mods. 

Here's how it looks right now:
image

### DIFF
--- a/src/main/java/spireMapOverhaul/SpireAnniversary6Mod.java
+++ b/src/main/java/spireMapOverhaul/SpireAnniversary6Mod.java
@@ -29,6 +29,7 @@ import com.megacrit.cardcrawl.unlock.UnlockTracker;
 import javassist.CtClass;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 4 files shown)


src/main/java/spireMapOverhaul/SpireAnniversary6Mod.java
```diff
@@ -208,6 +209,7 @@ public void receiveEditCards() {
     @Override
     public void receivePostInitialize() {
         initializedStrings = true;
+        unfilteredAllZones.forEach(AbstractZone::loadStrings);
```
modargo, line 212: As a funny consequence of how all the strings stuff is set up, the zones in `unfilteredAllZones` weren't getting their strings loaded before I added this. So if you wrote natural code that looked up a zone in this list and tried to get its name, it would be null, and you would be very confused (like I was). Best to not have traps like that.

Code like the map display worked due to calling `copy` first, since that made a new instance and loaded up strings again.

PR #16, opened 2023-12-21
 Add functions that can be overridden to prevent a zone from spawning in specific rows

modargo:
I hope that just reading the code gives a fairly clear idea of what's going on, so not going to add many comments to this PR.


src/main/java/spireMapOverhaul/SpireAnniversary6Mod.java
```diff
@@ -509,7 +509,7 @@ public void receivePostRender(SpriteBatch sb) {
     private static final float DROPDOWN_X = 400f;
     private static final float DROPDOWN_Y = 600f;
     private ModLabeledToggleButton filterCheckbox;
-    private static final float CHECKBOX_X = 650f;
+    private static final float CHECKBOX_X = 800f;
```
modargo, line 512: I made this change because the checkbox was hidden behind the dropdown for me. This is probably missing scaling or something like that.

It will probably need a better fix eventually but I didn't want to take the time to investigate right now.

PR #27, opened 2023-12-26
Candy Land

GameboyJ:
My first Zone of the anniversary mod, Candy Land.

Images for the cards will be added later, but wanted to create a pull request already to check if everything else if fine. If the consumables are approved, i'll make art for them. 

Candy Land changes card rewards to use Consumable cards, a subset of special colorless cards with rarities, that all are all fleeting, commons are simple, uncommons tend to come with a downside of some kind, and rares give a little bit of permanent progression.
In addition, the shop offers consumables instead of cards of your color, at half the price of colorless cards, and a Bite and Feed instead of colorless uncommon and rare cards.
All potions found in Candy Land are consumable (as in, not throwable), and so are the relics (as in strawberry, pear, mango or lee's waffle), and even the gold is consumable (chocolate coins that heal you instead of giving you gold). (this does not hold for chests as there was no hook for that)

Candy Land is there to help you while you are in there and shortly thereafter. Use it after or before some elites to heal back up or get some additional supplies for a boss. Candy Land is most beneficial if you can leverage the short-term bonusses for long-term bonusses, like helping you take additional elites.


src/main/java/spireMapOverhaul/zones/candyland/CandyLand.java
```diff
@@ -0,0 +1,197 @@
...
+
+    public void postAddButtons(ArrayList<AbstractCampfireOption> buttons) {
+        for (AbstractCampfireOption button : buttons){
+            if (button instanceof SmithOption){
+                button.usable = false;
+            }
+        }
+    }
+
+    public void postUseCampfireOption(AbstractCampfireOption option) {
```
modargo, line 123: You should change the rest campfire option text to say that it does this..

I don't think we have any hook for doing that, but it should be easy to patch. You can do it with a postfix patch for the constructor of `RestOption` that adds on to the `description` field.

src/main/java/spireMapOverhaul/zones/candyland/CandyLand.java
```diff
@@ -0,0 +1,197 @@
...
+        if(option instanceof RestOption) {
+            AbstractDungeon.player.increaseMaxHp(5, true);
+        }
+    }
+
+    public void postCreateShopCards(ArrayList<AbstractCard> coloredCards, ArrayList<AbstractCard> colorlessCards) {
```
modargo, line 129: You may want to prevent Gold Candy from spawning in the shop, since it will usually cost more gold that it gives.

All the times it will be a dead card aren't worth the tiny minority of times when you have Membership Card and might actually want to buy it.

src/main/java/spireMapOverhaul/zones/candyland/CandyLand.java
```diff
@@ -0,0 +1,197 @@
...
+        colorlessCards.clear();
+        colorlessCards.add(new Bite());
+        colorlessCards.add(new Feed());
+    }
+
+    public float modifyCardBaseCost(AbstractCard c, float baseCost) {
+        switch(c.cardID) {
+            case Bite.ID: return 80 + AbstractDungeon.merchantRng.random(1, 19); // Uncommon Colorless cost
+            case Feed.ID: return baseCost;
+            default: return (float) (baseCost*0.5);
```
modargo, line 153: This seems pretty reasonable to me. I am a bit worried that people might not want to buy `Fleeting` cards even at half price, but we can see what people's feedback for this zone is.

The easiest cards to evaluate for me are the ones that give permanent benefits (Banana, Chocolate Bar, Cocktail), and the price for those doesn't seem too far off. Roughly 75 gold for 2 (3) max HP, 4 (5) regen, or a random drinkable potion. That might be a bit too much given the comparison to Strawberry and just buying potions, so maybe it makes sense to start out with 40% of the normal cost instead of 50%. That would take these down to 60 gold.

 are all worth considering at this price, at least. ~75 gold for 2 (3) max HP, 4 (5) regen, or a random drinkable potion.

modargo (commented):
Congratulations on making the first zone PR! I've taken an initial pass and added comments. Most should be pretty straightforward to address, a few might require some discussion.

This zone seems reasonable in terms of design and balance -- strong healing and some max HP, in return for not getting other rewards. Something to aim for if you're low on HP, which adds meaningful decision-making to your pathing, which is good. I've made some small balance suggestions, but I expect there to be a lot of time for tweaks later once we actually start playtesting.

Since you made this PR, zone icons have been added in, so you should add one of those. The readme gives a quick overview and the other zones have examples (you'll need to pull in the most recent code).

Since this is the first zone and we're still getting used to the PR process for it, Gk and/or I may do another pass of this once everything else is in place. Sometimes we overlook things for new projects (it took a bit to get used to reviewing packs for Packmaster too).

src/main/java/spireMapOverhaul/zones/candyland/consumables/AbstractConsumable.java
```diff
@@ -0,0 +1,22 @@
...
+import com.megacrit.cardcrawl.rooms.MonsterRoom;
+import spireMapOverhaul.abstracts.AbstractSMOCard;
+
+
+abstract public class AbstractConsumable extends AbstractSMOCard {
+    private final CardRarity realRarity;
+
+    public AbstractConsumable(String cardID, int cost, CardType type, CardRarity rarity, CardTarget target) {
+        super(cardID, cost, type, CardRarity.SPECIAL, target);
+        realRarity = rarity;
```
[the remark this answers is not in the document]
GameboyJ: So, i do this so that they don't show up like normal colorless cards, but still have individual rarities. This is because I replace common cards with common consumables, uncommon cards with uncommon consumables and rare cards with rare consumables. I thought this would be the easiest way to keep track of that, without just hardcoding it in where it matters. Also, this way they get the correct rarity banners. 
The rarityies also have a theme, commons being just simple cards, uncommons having some form of downside, and rares giving some perma scaling.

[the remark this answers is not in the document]
modargo: Okay, then the issue that brings up is that as it stands, the rarities disappear when you save and reload (because they only get set when you see these cards in a shop or card reward). That's enough of a bug that something should change about this.

Here are two good options I see for resolving this:
* You could make the cards Special rarity and use `setDisplayRarity` to have them display the rarity banners
** This is probably the simplest and most straightforward approach, with the one downside being that any rarity-dependent effects won't work like people might expect (e.g. the Intrigue pack from Packmaster)
* You could make the cards Common/Uncommon/Rare but add `@NoPools` to leave them out of the normal card pool
** This is also fairly simple, with the one downside that effects that generate a random card of any color might generate these cards

There are probably other approaches too, but these came to mind since they're both nice and straightforward.

src/main/java/spireMapOverhaul/zones/candyland/consumables/rare/Banana.java
```diff
@@ -0,0 +1,29 @@
...
+import static spireMapOverhaul.SpireAnniversary6Mod.makeID;
+import static spireMapOverhaul.util.Wiz.atb;
+
+
+public class Banana extends AbstractConsumable {
+    public final static String ID = makeID(Banana.class.getSimpleName());
+
+    public Banana() {
+        super(ID, 0, CardType.SKILL, CardRarity.RARE, CardTarget.SELF);
+        baseMagicNumber = magicNumber = 2;
```
modargo, line 17: Discussing the rarity stuff reminded me -- you should add `tags.add(CardTags.HEALING);` to all the cards that give a permanent benefit (the four rares).

src/main/java/spireMapOverhaul/zones/candyland/CandyLand.java
GameboyJ: Actually, it turns out you can use postAddButtons to get the button for RestOption, then use reflection hacks to get and set the description. I'll update the documentation to mention changing descriptions when using postUseCampfireOption.

src/main/java/spireMapOverhaul/zones/candyland/CandyLand.java
GameboyJ: I could also increase the gold gained from Gold Candy, together with reducing the prices, it can be an investment. It would have to be more than 60 gold though, which seems like too much. Comparing Mango to Old Coin, 1 max HP is 25 gold, so to make it as strong as banana, it would have to be 50(75) gold

src/main/java/spireMapOverhaul/zones/candyland/consumables/AbstractConsumable.java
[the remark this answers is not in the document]
GameboyJ: oh, i was not aware of @NoPools, i think i will use that

modargo:
Looks good!

PR #33, opened 2023-12-30
Fix zone outline shader not rendering on Mac.

tetraminus:
This may need some testing on Windows and Linux, but this should fix an issue where zone outlines do not render when on a Mac. If needed, we may have to add some way to swap between shaders depending on platform. for now I'm hoping this will work out of the box, and I'll be able to test it soon. Also added a log for the shader just in case we run into an issue like this again. 
image


modargo:
Thanks. I confirmed that this at least works on Windows still by running the new code myself. We'll see if further changes are needed or not.

PR #35, opened 2023-12-31
Zone Potions Patch

bkerns724:
In addition to the potion patch I previously mentioned, I am partway through the content for my zone.
ZonePotions


erasels:
Could you separate the potions patches from your zone? Would allow landing it while letting you continue to work on your zone

bkerns724:
I need to work out some problems with git before I do pull requests

PR #41, opened 2024-01-02
Added junkyard basics

Dayvig:
Contains a boolean cannotSkipRewards that other zones can use if they choose to also not skip card rewards for whatever reason


erasels:
If your junk zone is not finished yet, please don't include it in this PR and just have the reward skip in it.

Also, please change the reward skip to be part of the reward modifying zone interface. You can do this by adding a default boolean allowRewardSkipping and returning true, then checking for that in your patch. That way it's easy to find

PR #42, opened 2024-01-02
Config menu using tootip

Pandemonium14:
If I keep up like this I'll have the highest number of PR merges


modargo:
You underestimate the amount of small PRs I create :). I follow my own words about them being better than big PRs.

PR #43, opened 2024-01-02
Broken Space Zone

tetraminus:
Broken Space adds a prismatic-like effect to card rewards, another card reward to combats, and replaces elite relic drops with special "Broken relics." It might need some balance touchups, but all the major functionality is there.


src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
```diff
@@ -0,0 +1,254 @@
...
+
+        AbstractCard c = CardLibrary.getAnyColorCard(rarity).makeCopy();
+
+        return c;
+
+
+    }
+
+    @Override
+    public ArrayList<AbstractCard> getAdditionalCardReward() {
```
erasels, line 146: Just a heads up, this does not include the upgrade logic from base game so these cards will never be upgraded. Even by the egg relics

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
```diff
@@ -0,0 +1,254 @@
...
+    public void modifyReward(RewardItem rewardItem) {
+        if (rewardItem.type == RewardItem.RewardType.RELIC) {
+            AbstractRelic origRelic = rewardItem.relic;
+            AbstractRelic newRelic = getValidBrokenRelic();
+
+            if (newRelic != null) {
+                rewardItem.relic = newRelic;
+                rewardItem.text = newRelic.name;
+
+                switch (origRelic.tier) {
```
erasels, line 173: Hmm, this might crash when the pool is empty and circlets show up

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
```diff
@@ -0,0 +1,254 @@
...
+        ArrayList<String> validRelics = new ArrayList<>();
+        for (String relicID : BrokenRelics) {
+            if (!adp().hasRelic(relicID)) {
+                validRelics.add(relicID);
+            }
+        }
+        if (validRelics.isEmpty()) {
+            return new Circlet();
+        }
+        AbstractRelic r;
```
erasels, line 209: Why not just check if the relic can spawn while checking if the player has the relic?

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
```diff
@@ -0,0 +1,254 @@
...
+        return r;
+
+    }
+
+    @Override
+    public void modifyRewardCards(ArrayList<AbstractCard> cards) {
+        int amount = cards.size();
+        cards.clear();
+
+        for (int i = 0; i < amount; i++) {
```
erasels, line 229: Same issue, upgrade and relic logic is skipped if you do this

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
```diff
@@ -0,0 +1,254 @@
...
+            clz = AbstractCard.class,
+            method = SpirePatch.CLASS
+    )
+    public static class UnnaturalCardField {
+        public static SpireField<Boolean> unnatural = new SpireField<>(() -> false);
+    }
+
+
+    public static void addBrokenRelic(String relicID) {
+        if (!BrokenRelics.contains(relicID)) {
```
erasels, line 248: You doN#t have to do this, but you could use a set for brokenRelics, sets have the property that all the items in them are unique so if you add an item twice, it will discard the second copy

src/main/java/spireMapOverhaul/zones/brokenSpace/DrawCardWithCallbackAction.java
```diff
@@ -0,0 +1,155 @@
...
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.vfx.PlayerTurnEffect;
+
+import java.util.ArrayList;
+import java.util.function.Consumer;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class DrawCardWithCallbackAction extends AbstractGameAction {
```
erasels, line 25: Unneeded, DrawCardAction can already do this, check out Scrape for an example. Also, this class checks for a harcoded handsize of 10 when it should check for BaseMod.MAX_HAND_SIZE. This is why you should always try to avoid copy pasting decompiled classes.

src/main/java/spireMapOverhaul/zones/brokenSpace/cardmods/UnreadableCardMod.java
```diff
@@ -0,0 +1,64 @@
...
+
+            if (Math.random() < 0.1) {
+                builder.append(" NL ");
+            }
+        }
+        return builder.toString();
+    }
+
+
+    // remove when it exits hand
```
erasels, line 57: Is this a TODO?

src/main/java/spireMapOverhaul/zones/brokenSpace/patches/BossRelicPricePatch.java
```diff
@@ -0,0 +1,25 @@
...
+        clz = AbstractRelic.class,
+        method = "getPrice"
+)
+public class BossRelicPricePatch {
+
+
+    @SpirePrefixPatch
+    public static SpireReturn<Integer> Prefix(AbstractRelic relic) {
+        // 14
+        if (Objects.requireNonNull(relic.tier) == AbstractRelic.RelicTier.BOSS) {
```
erasels, line 20: This might crash for circlets or wacky modded relics, also is this even still needed? I think you removed this?

src/main/java/spireMapOverhaul/zones/brokenSpace/patches/BrokenSpaceRenderPatch.java
```diff
@@ -0,0 +1,196 @@
...
+        return __instance.color != adp().getCardColor() && BrokenSpaceZone.UnnaturalCardField.unnatural.get(__instance);
+    }
+
+
+    private static boolean inBrokenSpace() {
+        return ZonePatches.currentZone() != null && ZonePatches.currentZone().id.equals("BrokenSpace");
+    }
+
+
+    @SpirePatch(
```
erasels, line 162: There is a scubscriber for the post render and this might be needed for other shaders as well, so could play please just add this to the main mod class instead? (SpireAnniversary6Mod)

Packmaster does the same thing

src/main/java/spireMapOverhaul/zones/brokenSpace/patches/BrokenSpaceRenderPatch.java
```diff
@@ -0,0 +1,196 @@
...
+
+    static {
+        brokenSpaceShader = new ShaderProgram(Gdx.files.internal(makeShaderPath("BrokenSpace/Glitch.vs")), Gdx.files.internal(makeShaderPath( "BrokenSpace/Glitch.fs")));
+        if (!brokenSpaceShader.isCompiled()) {
+            logger.warning("Broken Space shader: " + brokenSpaceShader.getLog());
+        }
+        brokenSpaceShader.begin();
+
+
+        fbo = new FrameBuffer(Pixmap.Format.RGBA8888, Gdx.graphics.getWidth(), Gdx.graphics.getHeight(), false, false);
```
erasels, line 194: This framebuffer size might be a bit overkill? Not that good with framebuffers and shaders so maybe i'm wrong, but generally making a fullscreen framebuffer is not very optimal

src/main/java/spireMapOverhaul/zones/brokenSpace/patches/CallingBellFix.java
```diff
@@ -0,0 +1,79 @@
...
+import com.megacrit.cardcrawl.ui.buttons.ProceedButton;
+import javassist.CannotCompileException;
+import javassist.CtBehavior;
+import jdk.vm.ci.code.site.Call;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.ArrayList;
+
+public class CallingBellFix {
```
erasels, line 19: Unneeded since boss relics were removed

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenDamaru.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public boolean canSpawn() {
+        return adp().chosenClass == AbstractPlayer.PlayerClass.WATCHER;
+    }
+
+    @Override
+    public void atBattleStart() {
+        addToBot(new ChangeStanceAction(new DivinityStance()));
```
erasels, line 49: Holy shit, broken indeed

erasels (changes requested):

src/main/resources/anniv6Resources/shaders/BrokenSpace/Glitch.fs
```diff
@@ -0,0 +1,115 @@
+
```
modargo, line 1: I checked this PR out locally because I wanted to see what some of the relics do, and the shaders that are applied to the relics need to be addressed. I'm not particularly sensitive to color/motion and I found having a relic using the shaders (specifically Broken Snecko Eye) on screen for a combat to be really obnoxious, almost nauseating. I would hate having that in my relic bar for a full run; I'd probably abandon the run or console remove the relic. 

Could you tone down this effect? I would suggest entirely or almost entirely eliminating the movement and color changes. You can have a distorted image of the relic, that _very slowly_ shifts color and changes how its distorted. If it constantly draws the eye to it because it's obviously moving and changing a way that nothing else on screen is, that's too distracting.

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,99 @@
+{
```
modargo, line 1: I'm going to give some thoughts on how I see the balance and play experience for these relics.

One thing to keep in mind with this is that if there's a relic people almost never want to pick up, or that leads to a negative experience for a lot of players, it's probably best to leave it out of the mod. Relics having major upsides and downsides is perfectly fine, but when things tip over into downsides being so large that the upside almost never justifies it or it feels bad, that's no good. Going too far in the other direction (i.e. very overpowered relics) is also bad).

That gets to the heart of my overall feedback for these relics: most of them go too far on the upsides and/or downsides, and would be better with less extreme effects (which can still be really impactful). In a number of cases this just means number tweaks or slight changes, but some relics might need a redesign. I do think there are a good number of cool ideas here, and this feedback is intended to help figure out how to execute the cool ideas in a way that players will appreciate, alongside identifying ideas that are better off being reworked so that people will appreciate what your zone does.

So with all that said, here are my thoughts, relic by relic:
* Broken Data Disk: This feels like it will too often be really bad or really good. -3 focus is enough that by default, Lightning and Frost Orbs will do nothing for their passive. But if you get back to positive focus with powers or clearing the debuff, you can quickly become overpowered. It's a bit too swingy. Maybe something like giving -1 focus in return for doubling the first orb channeled each turn would be sensible?

* Broken Snecko Skull: This is probably fine. There's some broken combinations with this (like Intangible), but the average power level seems okay.

* Broken Red Skull: This is probably fine. It might be too good at letting you stay ahead when you're already in good shape, but then again dexterity isn't that good for Ironclad. Easy enough to adjust later if needed.

* Broken Omamori: It looks like the description of this relic says it upgrades 1 card per rest, but the code upgrades 2 cards per rest. I'm not sure which is intended, but 3 curses is a lot, so I don't think I'd ever take this if it only upgraded 1 card. At 2 cards upgraded per rest, maybe... but 3 curses is such a short-term disadvantage that I still might not. Toning down the numbers might make this better.

* Broken Damaru: This seems insanely strong. Starting combat in Divinity will trivialize hallway fights and even some elites! Getting this relic is like triple Lantern just for the energy gain, and the 3x damage doesn't even have another relic you can compare it to. A different design might be best.

* Broken Wing Boots: The player won't have a clear idea of what this does and I'm not sure it's worth taking, which means this probably isn't a good idea. Best to find a different design.

* Broken Sozu: This effect is weird enough that I'm not at all sure about it. It fits the pattern of being too swingy, though; if you get this close to the end of the game, there's almost no downside. I suspect this would be better off with a less extreme design.

* Broken Snecko Eye: Setting aside whether this relic is good or bad, I'm concerned that it isn't fun to play with. Drawing cards that are almost completely disguised means either guessing blindly or counting cards to work out what they are (by looking at your hand/draw/discard to eliminate known cards until all that's left is the unreadable ones). Guessing blindly feels bad, counting cards is a play pattern we don't want to encourage. I'm also concerned about the unreadable card effect; having the card text constantly changing is no good. This one might need a rework.

* Broken Maw Bank: Seems reasonable. 500 gold might be a bit too much compared to Old Coin; the downside of needing to spend it all won't be a big deal a lot of the time (just buying out relics and potions and removal can easily be 800+ gold happily spent). Maybe try this at 400 gold, it's easy to tweak if it feels too weak.

* Broken Prayer Wheel: The downside here is excessive. Getting no card rewards is such a large drawback that taking this will doom most runs. Even with the 4x gold, unless you have Courier you're not likely to be able to buy enough cards from shops to make a deck that can win at high ascension (there's a reason Busted Crown is one of the worst boss relics). I think this would be a more interesting and balanced relic if it reduced the number of cards in rewards by 1, and had a lower multiplier for gold (e.g. 2).

* Broken Matryoshka: This relic has two issues. First, while the idea is reasonable, 5 relics for 2 empty chests is too good. Second, this relic is a forced skip if it drops after the act 1 chest for anyone planning to fight the Heart (since there's only two guaranteed chests left at that point, and if both are empty you have no way to get the blue key). Not everyone goes for the Heart fight, but a relic that a substantial subset of players almost always skips is no good. However, this can be fixed, as follows.
  * For the first problem, something like 3 relics for 2 empty chests seems balanced (immediate benefit now, neutral on relics in the long run, miss out on the gold that might be in the chests). Alternately, you could make it 4 relics, but always 3 commons and 1 uncommon (which matches Matryoshka giving a 75% chance of common and 25% chance of uncommon).
  * For the second problem, you would need to have one of the relics show up linked with the Blue Key (and say that in the relic description so that people going for the Heart understand they can take this relic).

* Broken Sundial: This seems like a relic you would almost always skip. Shuffling your discard pile back into your draw pile constantly puts you at risk of never finding your key cards in elite and boss fights. I recognize that like some of the other relics here, there's very specific synergies that could make this work, but that's not enough if the normal reaction is closer to "that's bad, skip". Might be best off finding a new design.

tetraminus (commented):
These arent replies for some reason, I've never used GitHub reviews before. either way, I'm working on fixing up and rebalancing.

src/main/resources/anniv6Resources/shaders/BrokenSpace/Glitch.fs
modargo: The relic shader is better now, but still very distracting.

Could you please make a no movement version? I understand that your intent with this is to making something interesting and cool, but seeing this in action, I've started to think that it's best to avoid having things in your relic bar move at all.

To clarify where we might be interpreting things differently, when I suggested having it move "very slowly" in my previous comment, I meant "so slow that it is almost imperceptible". Flipping around every few seconds is still too much. Thus the ask for a no movement version; that's a clear baseline we can start from. You can have the shader distort the base image by an amount that doesn't change, or only changes on save/reload or something like that.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenPrayerWheel.java
```diff
@@ -0,0 +1,49 @@
...
+    public int changeNumberOfCardsInReward(int numberOfCards) {
+        return numberOfCards - AMOUNT;
+    }
+
+    @Override
+    public String getUpdatedDescription() {
+        return DESCRIPTIONS[0] + AMOUNT + DESCRIPTIONS[1] + GOLD + DESCRIPTIONS[2];
+    }
+
+    @SpirePatch2(clz = RewardItem.class, method = "applyGoldBonus")
```
modargo, line 33: This patch doubles the gold when `applyGoldBonus` is called, but that method can be called multiple times -- it's called both in the constructor and by the `incrementGold` method. This means that when someone else writes logic to add extra gold, your patch will double the gold an extra time.

You can fix this by changing this to a patch targeting the two RewardItem constructors that take in a `gold` parameter. You should be able to use `@ByRef` to change the `gold` variable.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenSingingBowl.java
```diff
@@ -0,0 +1,46 @@
...
+                upgradableCards.add(c);// 103
+            }
+        }
+
+        for (int i = 0; i < AMOUNT; i++) {
+            if (!upgradableCards.isEmpty()) {
+                AbstractCard c = upgradableCards.remove(AbstractDungeon.cardRandomRng.random(upgradableCards.size() - 1));
+                c.upgrade();
+                AbstractDungeon.player.bottledCardUpgradeCheck(c);
+                AbstractDungeon.effectList.add(new ShowCardBrieflyEffect(c.makeStatEquivalentCopy(), (float) Settings.WIDTH/2, (float) Settings.HEIGHT/2));// 105 107 108 109
```
modargo, line 35: This should use `topLevelEffectsQueue` so that it shows properly when you're on the rewards screen.

You should also add the standard upgrade effect that gets used everywhere:
```
AbstractDungeon.topLevelEffectsQueue.add(new UpgradeShineEffect((float)Settings.WIDTH / 2.0F, (float)Settings.HEIGHT / 2.0F));
```

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenWingBoots.java
```diff
@@ -0,0 +1,63 @@
...
+import javassist.CannotCompileException;
+import javassist.expr.ExprEditor;
+import javassist.expr.FieldAccess;
+import javassist.expr.NewExpr;
+import spireMapOverhaul.BetterMapGenerator;
+import spireMapOverhaul.patches.BetterMapGenPatch;
+
+import static spireMapOverhaul.SpireAnniversary6Mod.makeID;
+
+public class BrokenWingBoots extends BrokenRelic {
```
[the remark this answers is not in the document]
modargo: The new effect also has the potential to cause trouble. This code is doing some weird stuff and reading through it made me pretty sure there would be bugs. I do see that it works for most events and event mostly works for combat events, which is actually a very good starting point.

However, I tried out the following as my first test case:
* Set this to choose between a random event and Masked Bandits
* Picked Masked Bandits
* Won the fight
* Saved and reloaded

This resulted in a crash. You should be able to reproduce it yourself.

It may be that this code, janky as it is, is pretty close to working in all cases and I just happened to find the one thing that needs to be fixed. Then again, it may be that there are other lurking issues and we just haven't found them yet (especially when you consider modded events and other stuff mods can do). This kind of thing can be a maintainability challenge.

I'll let @erasels weigh in with his thoughts as well, since I can see arguments for both options (trying to test and fix this vs. deciding that it's not worth the potential issues).

P.S.: For what it's worth, there are working examples of how to do this kind of "pick between events", though the one I know of is a single event that does it arther than a relic.

src/main/java/spireMapOverhaul/zones/brokenSpace/cardmods/UnreadableCardMod.java
```diff
@@ -0,0 +1,69 @@
...
+    public static final String ID = makeID("UnreadableCardMod");
+    private String name;
+
+    @Override
+    public void onInitialApplication(AbstractCard card) {
+
+        name = randomText(card.name.length());
+        card.exhaust = true;
+        card.isEthereal = true;
+        card.purgeOnUse = true;
```
modargo, line 18: You shouldn't use the `purgeOnUse` field. It has strange interactions, such as causing the card to not trigger Burst/Echo Form/etc.

Instead, you can use StsLib's purge modifier: `PurgeField.purge.set(card, true)`.

Also, if you set the card to purge, having it exhaust is redundant; purge overrides exhaust (the card will not go to the exhaust pile and it will not trigger effects like Feel No Pain). You can get rid of the exhaust part of this and change the relic description to say that it purges.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenSneckoEye.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public BrokenSneckoEye() {
+        super(ID, RelicTier.SPECIAL, LandingSound.CLINK, SneckoEye.ID);
+
+    }
+
+    @Override
+    public void atTurnStart() {
+        for (int i = 0; i < AMOUNT; i++) {
+            AbstractCard c = adp().masterDeck.getRandomCard(true).makeStatEquivalentCopy();
```
modargo, line 23: The new effect for this is much better (and the unreadable mod no longer having constantly changing text is much more pleasant). One unreadable card per turn that you can decide whether you want to risk playing seems fine.

One thing that needs to be changed about this -- sources of card generation like this need to avoid healing cards (anything where `card.hasTag(CardTags.HEALING)` returns true). You should add in logic to prevent those cards from being generated.

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,105 @@
...
+      " [E] to use. When you spend energy on a potion, use another random potion.",
+      " Energy"
+    ]
+  },
+  "${ModID}:BrokenSneckoEye": {
+    "NAME": "Broken Snecko Eye",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "At the start of each turn, add #b",
+      " copy of a random card from your master deck into your hand. It #yExhausts, is #yEthereal, and is unreadable."
```
modargo, line 67: This can just say "deck" ("master deck" is something modders tend to say because that's that's the field you access in the code, `AbstractPlayer.masterDeck`).

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenDataDisk.java
```diff
@@ -0,0 +1,66 @@
...
+    public boolean shouldTrigger = true;
+    private boolean usedThisTurn = false;
+
+    public BrokenDataDisk() {
+        super(ID, RelicTier.SPECIAL, LandingSound.CLINK, DataDisk.ID);
+    }
+
+    @Override
+    public void onChannel(AbstractOrb abstractOrb) {
+        if (!shouldTrigger || usedThisTurn) {
```
modargo, line 27: What's the reason you need both `shouldTrigger` and `usedThisTurn`? Looks to me like only having `usedThisTurn` would be enough.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenDataDisk.java
```diff
@@ -0,0 +1,66 @@
...
+            return;
+        }
+
```
modargo, line 30: You should add a `this.flash();` since that's the standard way to show a relic activating based on something the player did.

modargo: You should make sure your other relics have a `flash()` call at appropriate times too. It looks like a few do, but mostly where you took code from the base game (e.g. Broken Matryoshka on chest open).

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,105 @@
...
+    "DESCRIPTIONS": [
+      "When you would enter an event in a ? room, choose between two events."
+
+    ]
+  },
+  "${ModID}:BrokenSozu": {
+    "NAME": "Broken Sozu",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "In-Combat Potions cost #b",
```
modargo, line 57: Interesting effect. It seems reasonably balanced, though if it turns out to be too weak you could consider having it lose energy next turn instead of on the current turn. That would be a bit more forgiving to the player.

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
modargo: Relic balance seems reasonable after the changes (though Broken Wing Boots may need more discussion for other reasons)

Thanks for going through and responding to all the feedback!

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
modargo: For the text of this relic, seeing "1 [E]" is weird. Normally you don't have the number when it's a fixed small amount, so this should just be "In-Combat potions cost [E] to use."

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenMatryoshka.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public void onEquip() {
+        for (int i = 0; i < AMOUNT-1; i++) {
+            AbstractRelic r = AbstractDungeon.returnRandomScreenlessRelic(getRelicTier());
+            AbstractDungeon.getCurrRoom().rewards.add(new RewardItem(r));
+        }
+        AbstractRelic r = AbstractDungeon.returnRandomScreenlessRelic(getRelicTier());
+        RewardItem reward = new RewardItem(r);
+        AbstractDungeon.getCurrRoom().rewards.add(reward);
+        if (!Settings.hasSapphireKey){
```
modargo, line 31: This should also check `Settings.isFinalActAvailable` (you can see that `AbstractChest.open` checks that when it calls this same method).

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenPrayerWheel.java
modargo: I realized what I suggested might have a different bug -- I think it will double the amount of gold when you save and reload. So ignore the suggestion solution in my previous comment.

What you can do instead is patch `applyGoldBonus`, but have this work the same way other gold bonuses (e.g. Golden Idol) work. They use the `bonusGold` field to store the extra gold, which avoids issues with repeatedly calling `applyGoldBonus` (since the bonus amount is recalculated based on the normal amount of gold). Looks like that will need an insert patch, but shouldn't be that bad.

This is what I should have thought of first, really. There are other ways you could set this up, but it's often the best bet to write code to work the same way that the base game does similar functionality. It makes compatibility a lot easier.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenWingBoots.java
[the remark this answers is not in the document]
tetraminus: I just looked into that crash, it seems relatively easily fixable (famous last words). I'm going to dive into it a little more today.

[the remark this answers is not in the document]
tetraminus: nevermind? fixed it.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenPrayerWheel.java
```diff
@@ -0,0 +1,59 @@
...
+        public static void patch(RewardItem __instance, int tmp) {
+            if (AbstractDungeon.player.hasRelic(makeID(ID))) {
+                int prayerWheelAmt = 0;
+                for (AbstractRelic r : AbstractDungeon.player.relics) {
+                    if (r instanceof BrokenPrayerWheel) {
+                        prayerWheelAmt++;
+                    }
+                }
+
+                __instance.bonusGold += (int) (tmp * Math.pow(GOLD, prayerWheelAmt));
```
modargo, line 46: Using `Math.pow` here isn't seem right. If you had two copies of this relic. it would square the amount of gold. Three would cube it. Even at one copy I don't think this is right; if you get 100 gold, this will give you 200 bonus gold for a total of 300 gold.

Having multiple copies of a relic is a very niche interaction, but the simplest way to handle this is to avoid exponential scaling at all. Treat each copy of the relic as an additive increase (e.g. +100% gold), and write this as:
```suggestion
                __instance.bonusGold += (int) (tmp * prayerWheelAmount * (GOLD - 1)));
```

src/main/resources/anniv6Resources/shaders/BrokenSpace/Glitch.fs
modargo: Thanks, I'm much more of a fan of the new, non-moving version. The distortion effect is cool but the relic is still vaguely recognizable, and it's no longer so distracting.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenPrayerWheel.java
tetraminus: yeah, that makes sense. I totally misunderstood what that function was doing, Thanks!

src/main/resources/anniv6Resources/images/ui/zoneIcons/BrokenSpace.afphoto
modargo, line 1: Didn't we want to delete this file? With how .gitignore works, you might need to do this as a multi-step process:
* Remove the *.afphoto line from the .gitignore
* Delete this file
* Add the *.afphoto line back to the .gitignore

You might need to have each of those steps as its own commit, I'm not sure.

modargo:
This is looking in pretty good shape now. What's left is two things:
* I've asked Gk to weigh in on Broken Wing Boots, and if we decide to keep it in I'll want to look at the code more closely (the fix you made for the crash looks sensible, but there's a few other things I want to look at)
* Since the PR is complicated (mostly due to the relics), I plan to take another pass through things to see if there's anything else that jumps out

After both of those and any resulting changes are done, this should be ready to merge. I'm looking forward to experiencing people's zones in a real run!

src/main/resources/anniv6Resources/images/ui/zoneIcons/BrokenSpace.afphoto
tetraminus: should be gone now.

erasels:
I think the broken wing boots can stay, bit jank but it's a creative effect. If we get too many bug reports about it, I might just nuke it post release but we'll see. I fear certain event packs might interact badly with it.

Also, I see a lot of earlier comments have not been adressed, please don#t forget to do so!

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenSneckoSkull.java
```diff
@@ -0,0 +1,61 @@
...
+
+
+    @Override
+    public String getUpdatedDescription() {
+        return DESCRIPTIONS[0] + AMOUNT + DESCRIPTIONS[1];
+    }
+
+
+    @Override
+    public boolean onReceivePower(AbstractPower abstractPower, AbstractCreature abstractCreature) {
```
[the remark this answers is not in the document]
erasels: Nice, maybe just check if the power amount is > 0 to be safe since there are buffs that can have negative amounts or just have 0

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
```diff
@@ -0,0 +1,261 @@
...
+
+    }
+
+    @Override
+    public ArrayList<AbstractCard> getAdditionalCardReward() {
+        ArrayList<AbstractCard> cards = new ArrayList<>();
+        int amount = 3;
+
+        for (AbstractRelic r : AbstractDungeon.player.relics) {
+            amount = r.changeNumberOfCardsInReward(amount);
```
modargo, line 163: After pulling in the most recent code, you can replace this logic with `int amount = this.getNumberOfCardsInReward();`. Like with the upgrade stuff, we made a helper function for this after seeing this PR.

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
modargo: Useful information: Circlet's are `RelicTier.SPECIAL`, which is the rarity also used for event relics.

I suggest adding the following below:
```
case SPECIAL:
    // There's no pool for special relics, so we don't need to do anything
    break;
```

I think it's fine to do nothing for Circlets (and for any other special rarity relic that somehow shows up as an elite drop), but it's good to show that it's been specifically thought about, and I wouldn't want the log statement in the `default` case to show up.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenDataDisk.java
```diff
@@ -0,0 +1,64 @@
...
+    public BrokenDataDisk() {
+        super(ID, RelicTier.SPECIAL, LandingSound.CLINK, DataDisk.ID);
+    }
+
+    @Override
+    public void onChannel(AbstractOrb abstractOrb) {
+        if (usedThisTurn) {
+            return;
+        }
+        addToTop(new AbstractGameAction() {// 29
```
modargo, line 29: Looks like this action is doing nothing now and you can delete it.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenSneckoEye.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public BrokenSneckoEye() {
+        super(ID, RelicTier.SPECIAL, LandingSound.CLINK, SneckoEye.ID);
+
+    }
+
+    @Override
+    public void atTurnStart() {
+        flash();
+        for (int i = 0; i < AMOUNT; i++) {
```
modargo, line 23: It's very very niche, but seeing code like this makes me think of an obvious edge case: when the player's deck has only healing cards.

Let's add the following to cover that (at the start of this function):
```
        if (adp().masterDeck.group.stream().allMatch(c -> c.hasTag(AbstractCard.CardTags.HEALING))) {
            return;
        }
```

There are other ways to handle this, but the code I gave seems like the simplest approach that fits in with the current code structure.

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
+{
+  "${ModID}:BrokenDataDisk": {
+    "NAME": "Broken Data Disk",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "At the start of each combat, lose #b",
+      " #yFocus. The first time you channel an orb each turn, channel another orb of the same type."
```
modargo, line 7: I'm taking a look at minor stuff like wording consistency. Channel is a keyword and the base game consistently capitalizes Orb.
```suggestion
      " #yFocus. The first time you #yChannel an Orb each turn, #yChannel another Orb of the same type."
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+      "% #yHP, you have #b",
+      " extra #yDexterity."
+    ]
+  },
+  "${ModID}:BrokenOmamori": {
+    "NAME": "Broken Omamori",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "Add #b",
+      " #yCurses to your deck. Whenever you rest, upgrade #b",
```
modargo, line 32: Normally the base game uses red for the word Curses in relic descriptions (see Omamori and Du-Vu Doll). Also, upgrade is a keyword.
```suggestion
      " #rCurses to your deck. Whenever you rest, #yUpgrade #b",
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+      " random cards in your deck."
+    ]
+  },
+  "${ModID}:BrokenDamaru": {
+    "NAME": "Broken Damaru",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "At the end of your turn, you may #yExhaust a card in your hand. Gain #yMantra equal to it's cost.",
```
modargo, line 40: it's -> its
```suggestion
      "At the end of your turn, you may #yExhaust a card in your hand. Gain #yMantra equal to its cost.",
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+      "Select a card to exhaust for mantra."
```
modargo, line 41: Probably best to capitalize Mantra since it's a keyword (though card select screen like this are basically the wild west for text consistency).
```suggestion
      "Select a card to exhaust for Mantra."
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+
+    ]
+  },
+  "${ModID}:BrokenWingBoots": {
+    "NAME": "Broken Wing Boots",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "When you would enter an event in a ? room, choose between two events."
```
modargo, line 49: Juzu Bracelet uses `#y?`.
```suggestion
      "When you would enter an event in a #y? room, choose between two events."
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+
+    ]
+  },
+  "${ModID}:BrokenSozu": {
+    "NAME": "Broken Sozu",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "In-Combat Potions cost [E] to use. When you spend energy on a potion, use another random potion.",
```
modargo, line 57: No need to capitalize the first couple words here.
```suggestion
      "In-combat potions cost [E] to use. When you spend energy on a potion, use another random potion.",
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+      " #yGold next time you enter a shop, but lose #b",
+      " #yGold when you leave."
+    ]
+  },
+  "${ModID}:BrokenPrayerWheel": {
+    "NAME": "Broken Prayer Wheel",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "Card rewards contain #b",
+      " less card, but each gold reward is multiplied by #b",
```
modargo, line 83:
```suggestion
      " less card, but each #yGold reward is multiplied by #b",
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+      "."
+    ]
+  },
+  "${ModID}:BrokenMatryoshka": {
+    "NAME": "Broken Matryoshka",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "Add #b",
+      " Relics to the rewards, but the next #b",
```
modargo, line 92:
```suggestion
      " #yRelics to the rewards, but the next #b",
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+      " chests you open are empty. If you don't have the Sapphire Key, it will be linked to a relic."
```
modargo, line 93:
```suggestion
      " chests you open are empty. If you don't have the Sapphire Key, it will be linked to a #yRelic."
```

src/main/resources/anniv6Resources/localization/eng/BrokenSpace/Relicstrings.json
```diff
@@ -0,0 +1,104 @@
...
+    ]
+  },
+  "${ModID}:BrokenSingingBowl": {
+    "NAME": "Broken Singing Bowl",
+    "FLAVOR": "no",
+    "DESCRIPTIONS": [
+      "Whenever you gain Max HP, upgrade #b",
```
modargo, line 100:
```suggestion
      "Whenever you gain Max HP, #yUpgrade #b",
```

modargo (commented):
I went through the code again and noted a few more things, plus looked at wording to catch any minor issues (it should be easy to accept the suggestions directly in Github's interface). After the remaining comments from me and Gk are addressed, I think this will be good to go.

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
tetraminus: pushing fix.

src/main/java/spireMapOverhaul/zones/brokenSpace/patches/BrokenSpaceRenderPatch.java
tetraminus: I think otherwise I would have to manually reposition the render locations. I may be wrong but this is certainly the simplest path.

src/main/java/spireMapOverhaul/zones/brokenSpace/relics/BrokenSneckoSkull.java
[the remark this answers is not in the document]
tetraminus: should be fixed next commit

src/main/java/spireMapOverhaul/zones/brokenSpace/BrokenSpaceZone.java
tetraminus: I might change it in the future I'm fine with this for now.

erasels:
I think that's it? Zone's ready to land now, right?

tetraminus:
I can always patch in more fixes if needed. Thanks for all the feedback!

erasels:
Thanks for being a good sport throught all these changes and for your contribution!

PR #44, opened 2024-01-03
Thieves' Hideout zone

modargo:
This PR has the full implementation of my Thieves' Hideout zone.
* Four remix fights (with thieves and slavers from the base game)
* Two fights with new monsters (Bandit Lieutenant and Taskmaster, Thief King and lackeys)

The two fights are both part of a special node in the zone that con-+tains a forced event. I did not use ModifiedEventRateZone for two reasons: (1) that doesn't guarantee that an event node is an event (rather than a fight or shop or treasure) and (2) it would have been more awkward to set it up to only affect a single specific node (I wanted one special node in the last row of the zone). Instead, I made `ForcedEventRoom`, which may be something other zones find useful, and I placed the room manually.

Since the rewards for the forced event are at the level of the Colosseum event (100 gold and two relics), there are no elite fights in the zone. Think of it concentrating all the elite fights that might be in the zone into one node. Like Colosseum, you also have the option of leaving after the first fight, but unlike Colosseum, you get 100 gold if you leave.

The first fight (Bandit Lieutenant and Taskmaster) aims to be around the difficulty of a typical elite fight. The second fight (Thief King, Looter, and Mugger) aims to be around the difficulty of a super elite fight. The current numbers are based on a few quick sample fights and tuning things against other act 2 elites, but once I start doing serious playtesting I'm sure I'll make various balance adjustments. I'm just waiting for more zones before doing much playtesting.

I've included screenshots of the two new fights and the Bestiary popups for the new monsters, plus the middle screen of the Thief King event and what the zone looks like on the map (you can see the flame effect behind the final ? node). Should give you an idea of what this zone is like at a glance.

The amount of code for this zone is not that large, so you should be able to look through all of it if you want. Shouldn't be anything crazy here so feel free to engage a lot or a little or anywhere in between.
thieves hideout 1
thieves hideout 2
thieves hideout 3
thieves hideout 4
thieves hideout 5
thieves hideout 6


src/main/java/spireMapOverhaul/zones/thieveshideout/ThievesHideoutZone.java
```diff
@@ -0,0 +1,137 @@
...
+                new AbstractMonster[] {
+                    new Looter(-270.0F, 15.0F),
+                    new SlaverRed(130.0F, 0.0F)
+                })),
+            new ZoneEncounter(MUGGER_AND_BLUE_SLAVER, 2, () -> new MonsterGroup(
+                new AbstractMonster[] {
+                    new Looter(-270.0F, 15.0F),
+                    new SlaverRed(130.0F, 0.0F)
+                })),
+            new ZoneEncounter(THREE_LOOTERS_WEAK, 2, () -> new MonsterGroup(
```
modargo, line 111: All the other fights seemed reasonably balanced for the hard pool purely with the mix of base game enemies I made, but I wanted a three Looters and that would normally be 33 damage on turn 1. That's a bit crazy for act 2.

Giving them all -2 strength (implemented in the `WeakLooter` class) makes it into 27 damage. That's high, but it's line with other dangerous act 2 fights in the hard pool. In particular, it's the same as Shelled Parasite and Fungi Beast on turn 1.

(All numbers are A20 values.)

src/main/java/spireMapOverhaul/zones/thieveshideout/monsters/BanditLieutenant.java
```diff
@@ -0,0 +1,141 @@
...
+import com.megacrit.cardcrawl.cards.status.Dazed;
+import com.megacrit.cardcrawl.core.CardCrawlGame;
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.localization.MonsterStrings;
+import com.megacrit.cardcrawl.powers.DrawReductionPower;
+import com.megacrit.cardcrawl.powers.VulnerablePower;
+import com.megacrit.cardcrawl.vfx.combat.SmokeBombEffect;
+import spireMapOverhaul.SpireAnniversary6Mod;
+
+public class BanditLieutenant extends CustomMonster {
```
modargo, line 22: I didn't use `AbstractSMOMonster` for these monsters (or the `PhasedEvent` stuff for the events) because I already have good setups for the older style of making monsters/events and don't find the new style any faster.

src/main/java/spireMapOverhaul/zones/thieveshideout/ThievesHideoutZone.java
```diff
@@ -0,0 +1,137 @@
...
+    private static final String LOOTER_AND_RED_SLAVER = SpireAnniversary6Mod.makeID("LOOTER_AND_RED_SLAVER");
+    private static final String MUGGER_AND_BLUE_SLAVER = SpireAnniversary6Mod.makeID("MUGGER_AND_BLUE_SLAVER");
+    private static final String THREE_LOOTERS_WEAK = SpireAnniversary6Mod.makeID("THREE_LOOTERS_WEAK");
+    public static final String BANDIT_LIEUTENANT_AND_TASKMASTER = SpireAnniversary6Mod.makeID("BANDIT_LIEUTENANT_AND_TASKMASTER");
+    public static final String THIEF_KING = SpireAnniversary6Mod.makeID("THIEF_KING");
+
+    public ThievesHideoutZone() {
+        super(ID, Icons.MONSTER);
+        this.width = 2;
+        this.height = 5;
```
modargo, line 41: I made this a narrow but long zone to have more of a feeling of going through a thief-heavy area before facing the Thief King at the end.

src/main/java/spireMapOverhaul/zones/thieveshideout/ThievesHideoutZone.java
```diff
@@ -0,0 +1,137 @@
...
+        //It also has normal fights that are tuned as hard pool fights, so that's another reason to prevent early spawns
+        return false;
+    }
+
+    @Override
+    public void manualRoomPlacement(Random rng) {
+        //We put the special node with the Thief King fight in the last row of the zone (that isn't a protected row)
+        List<MapRoomNode> possibleNodes = new ArrayList<>();
+        for (MapRoomNode node : this.nodes) {
+            if (node.getRoom() == null && !this.isProtectedRow(node.y)) {
```
modargo, line 71: Being able to write this check is why I made `AbstractZone.isProtectedRow` protected instead of private.

src/main/java/spireMapOverhaul/zones/thieveshideout/patches/HighlightForcedEventRoomPatch.java
```diff
@@ -0,0 +1,58 @@
...
+import com.badlogic.gdx.graphics.g2d.SpriteBatch;
+import com.badlogic.gdx.math.MathUtils;
+import com.evacipated.cardcrawl.modthespire.lib.*;
+import com.megacrit.cardcrawl.map.MapRoomNode;
+import com.megacrit.cardcrawl.vfx.FlameAnimationEffect;
+import spireMapOverhaul.zones.thieveshideout.rooms.ForcedEventRoom;
+
+import java.util.ArrayList;
+
+public class HighlightForcedEventRoomPatch {
```
modargo, line 14: This is basically just a patch copy of the burning elite vfx. I'd written something like this before so it was easy to adapt.

Happy to swap this out for something else if we decide on a standard like the `!` idea you mentioned.

erasels: Personally I think an exlamation mark has less room for misunderstandings since the burning effect is already associated with keys. However I don't care overmuch, if it's too much work we can just go with this

erasels (approved):
Looks good, I only flew over the monster code because I trust that you've tested them and the code itself isn't that complex,just expansive.

If you decide that that you don't want to make the ! for the forced event room, then feel free to merge.

modargo:
> Looks good, I only flew over the monster code because I trust that you've tested them and the code itself isn't that complex,just expansive.
> 
> If you decide that that you don't want to make the ! for the forced event room, then feel free to merge.

Yup, sounds good. I'll probably give myself a bit to think about it and come back to this tonight, to either merge or add the ! thing. Leaning towards merge but it doesn't hurt to let my thoughts settle.

PR #54, opened 2024-01-06
basegame method can crash

Alchyr:
there can be a null cardqueue.card item for the end of turn


erasels:
Added a fix to basemod, kio will merge it before this releases, trust 🙏🏻

PR #64, opened 2024-01-07
Add Monster zoo zone

erasels:
Adds a hook to change the number of cards in the card rewards to RewardModifyingZone.

But the main attraction is that it adds the Monster Zoo Zone that replaces all nodes within it with normal Monsterrooms (it excludes early rows in act 1, and the other special rows)

Monsters within have 2 more strength and when beat give you double gold, an additional card in the card reward and some other small tweaks. While within a monster room, semi transparent monstrs streak across the background to give you the feel that it's crowded.

image
The icon has changed a bit, to be simpler but it looks like this.


src/main/java/spireMapOverhaul/zones/monsterZoo/MonsterZooZone.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    // All monsters gain an amount of strength
+    @Override
+    public void atPreBattle() {
+        Wiz.forAllMonstersLiving(m -> Wiz.atb(new ApplyPowerAction(m, null, new StrengthPower(m, getStrAmt(m)))));
+    }
+
+    private int getStrAmt(AbstractMonster m) {
+        //Might make this dynamic as in 1 + actNum
```
modargo, line 116: Interestingly enough, I also landed on 2 strength as the amount to give to normal monsters in Corrupt the Spire, regardless of act. Having it vary by act number (like the burning elite buff) is reasonable too, but what I found was that:
* For act 1, 2 strength is a significant percent increase
* For act 2, hallway fights are challenging enough that 2 strength is a real additional challenge, even if its not as big of a percent increase 
* For act 3, hallway fights vary widely in difficulty, but 2 strength is still enough to make Darklings and Jaw Worms into big challenges with frontloaded damage

Also, one edge case to be aware of that _will_ make people rage -- the byrds fight. The impact strength has on that fight is so dramatically greater than every other fight that I special cased it for Corrupt the Spire to give half the amount.

src/main/java/spireMapOverhaul/zones/monsterZoo/BackgroundMonsterEffect.java
```diff
@@ -0,0 +1,373 @@
...
+import com.megacrit.cardcrawl.monsters.beyond.*;
+import com.megacrit.cardcrawl.monsters.city.*;
+import com.megacrit.cardcrawl.monsters.exordium.*;
+import com.megacrit.cardcrawl.vfx.AbstractGameEffect;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+
+//Mostly copied from https://github.com/a-personal-account/PaleOfTheAncients/blob/master/src/main/java/paleoftheancients/finarubossu/vfx/BackgroundMonster.java
```
modargo, line 28: Ah yes, I remember seeing this in the background of the final boss for Pale.

I skimmed through this class, not much need to review it in detail.

src/main/java/spireMapOverhaul/zones/monsterZoo/MonsterZooZone.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Override
+    public void onVictory() {
+        BackgroundMonsterEffect.masterDispose();
+    }
+
+    // All monsters gain an amount of strength
+    @Override
+    public void atPreBattle() {
+        Wiz.forAllMonstersLiving(m -> Wiz.atb(new ApplyPowerAction(m, null, new StrengthPower(m, getStrAmt(m)))));
```
modargo, line 112: Did you consider having a bigger set of potential buffs (that would be randomly picked from, either instead of or in addition to strength)? Not needed for merging this or anything, but if you want to build on the zone it's something you could add in. I think the extra variety would be interesting if monsters sometimes have metallicize or malleable or artifact.

Though at this point I'm just describing what Corrupt the Spire does with fights, which is a funny place to end up at.

src/main/java/spireMapOverhaul/zones/monsterZoo/BackgroundMonsterEffect.java
```diff
@@ -0,0 +1,373 @@
...
+            this.color.a -= 0.1F;
+        }
+        this.direction.x *= Settings.scale * 50F;
+        this.direction.y *= Settings.scale * 50F;
+
+        float multiplier = Settings.WIDTH / 2F / Math.abs(direction.x) * 1.3F;
+        this.ac.drawX = Settings.WIDTH / 2F - direction.x * multiplier;
+        this.ac.drawY =  MathUtils.random(Settings.HEIGHT * 0.25f, Settings.HEIGHT * 0.75f);
+
+        this.rotation = 0;
```
modargo, line 56: I see that the rotation velocity logic below is commented out, which I assume is to avoid the potential distraction of having enemy images spinning in the background as they go by.

I checked this out locally and while not having active rotation makes sense, it might be cool if the enemies had some rotation applied. If you don't want enemies to be upside down, you could just restrict the range of rotation so that the enemies are always still upright, just tilted.

modargo (approved):
Looks great. I wrote some minor comments that might lead to tweaks you want to make, but it would also be fine if you decided not to. Whichever you decide, this is ready to merge.

src/main/java/spireMapOverhaul/zones/monsterZoo/MonsterZooZone.java
erasels: oh, good point for byrds

erasels:
After some additional thinking, I decided that that I'm not happy with other modifiers being on monster. While a bit of randomness might be a nice change of pace, it feels like it would be frustrating more often than not.

Having a group of monsters get passive block would be a drawn out experience, something I don't really want to happen. Couldn't think of any good damage modifiers here either, Anger scaled too hard, Curiosity was too all or nothing, thorns were too limiting and so on.

PR #65, opened 2024-01-07
add heavenly clouds zone

Darkglade1:
The Heavenly Clouds zone gives all monsters inside 30% less HP but they get 3(4) Flight. Losing all Flight causes a turn of Stun and a turn of flightlessness. Card rewards in this zone are modified to have the Flight mod, which adds "Draw 1 card" to their effects.


erasels:
Code looks great, just asked Modargo to leave some balancing thoughts here because some interactions seems incredibly problematic with both the card modifier and the monsters this could be applied to.

He'll formulate his thoughts and leave them here once he's done.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public Color getColor() {
+        return Color.SKY.cpy();
+    }
+
+    @Override
+    public void atBattleStart() {
+        int flightAmount = 3;
```
modargo, line 41: Cool zone! Flight is a good combat modifier for giving things a twist. Players can even be strategic about entering or avoiding this zone based on how well their deck does against Flight.

For balance, I think it's important to have the amount of Flight given to enemies vary. The Byrds fight is challenging because its at the start of act 2 and has three enemies, so even if you manage enough attacks in a turn there are still two other Byrds hitting you. The further away from that you get, the less balanced these specific values are.

With this static amount of flight, I would expect single enemy fights to be substantially easier, and act 3 fights to also be substantially easier. Think about fighting Gremlin Nob or Lagavulin at low ascension, and being able to stun them with 3 attacks. That's something you can do in act 1 no problem. It's perfectly fine for this zone to affect fights unequally, but some tuning of these values would go a long way towards making the zone feel fair.

There's a variety of ways to address this, but the most straightforward in my mind is to have the amount of Flight vary by act, whether the enemy is an elite, and whether the fight has only one enemy. Here's what I'd propose:
* All single enemy fights get +1 (otherwise they're just too easy)
* All elites in act 2 and 3 get +1 (because elites should stay challenging)
* All enemies in act 3 get +1 (since by that point it's a lot easier to play many attacks in a turn)

Lots of other ways to set this up, but that's my first stab at values for this that I think will be reasonably balanced.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
```diff
@@ -0,0 +1,60 @@
...
+import spireMapOverhaul.abstracts.AbstractZone;
+import spireMapOverhaul.util.Wiz;
+import spireMapOverhaul.zoneInterfaces.CombatModifyingZone;
+import spireMapOverhaul.zoneInterfaces.RewardModifyingZone;
+
+import java.util.ArrayList;
+
+import static spireMapOverhaul.util.Wiz.atb;
+
+public class HeavenlyCloudsZone extends AbstractZone implements CombatModifyingZone, RewardModifyingZone {
```
modargo, line 18: Two edge cases are worth thinking about:
1. What happens to the Byrds fight? It would be pretty weird to see them have flight twice. Given how flight is their thing, special casing how this zone interacts with Byrds makes the most sense to me.
2. What about newly spawned enemies (e.g. Gremlin Leader's gremlins)? We don't have a hook for this yet, and it's fine to leave it out, but I wanted to make sure you were happy with how this zone worked with that.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/FlightMod.java
```diff
@@ -0,0 +1,43 @@
...
+    public static final String[] TEXT = uistrings.TEXT;
+
+    @Override
+    public AbstractCardModifier makeCopy() {
+        return new FlightMod();
+    }
+
+    @Override
+    public void onUse(AbstractCard card, AbstractCreature target, UseCardAction action) {
+        AbstractDungeon.actionManager.addToBottom(new DrawCardAction(1));
```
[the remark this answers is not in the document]
erasels: Hmm, an alternative would be to go the same way as eric and just make it work once.

Could go ever farther and make it draw 1 card when drawn and then the effect dissapates for that combat.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void atBattleStart() {
+        int flightAmount = 3;
+        if (AbstractDungeon.ascensionLevel >= 19) {
+            flightAmount = 4;
+        }
+        for (AbstractMonster mo : Wiz.getEnemies()) {
+            atb(new ApplyPowerAction(mo, mo, new HeavenlyFlightPower(mo, flightAmount), flightAmount));
+            mo.currentHealth = (int)((float)mo.maxHealth * HEALTH_MODIFIER);
```
modargo, line 47: If something causes enemies to start with lower current health than max health (e.g. Preserved Insect), this will wipe that out, right? I believe the zone hooks mostly happen after the equivalent relic hooks.

I think it would be better to apply the percentage health reduction to both current and max health, instead of setting current to a percentage of max.
```suggestion
            mo.currentHealth = (int)((float)mo.currentHealth * HEALTH_MODIFIER);
```

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyFlightPower.java
```diff
@@ -0,0 +1,83 @@
...
+        super(POWER_ID, NAME, AbstractPower.PowerType.BUFF, false, owner, amount);
+        this.loadRegion("flight");
+        this.priority = 50;
+        this.initialAmount = amount;
+        updateDescription();
+    }
+
+    @Override
+    public void onInitialApplication() {
+        owner.drawY += HEIGHT_ADJUSTMENT * Settings.scale;
```
modargo, line 37: This is great, I was wondering if it would make them float up a bit.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/FlightMod.java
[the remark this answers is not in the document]
modargo: Only working once (either when played on when drawn) is a great suggestion. I'd be on board with that too.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
erasels: Act 3 elites getting 5 flight sounds terrifying that has to be tested. Playing that many attacks is hard for many decks and realsitically could lead to a lot of deaths and making this zone an instant avoid

Darkglade1: I definitely agree that single enemy encounters should get +1 flight; I'm not so sure about the other two additions since I agree with Gk that having these effects potentially stack up to 5 or more Flight would likely be problematic. Unfortunately the mechanic is rather volatile to balance since the addition of +/-1 Flight could be the difference between wiping to an encounter or beating it with ease 😛 

I think in this instance, I would rather make encounters too easy instead of too hard since that would likely provide a better experience for the player on average.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/FlightMod.java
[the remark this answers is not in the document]
Darkglade1: Great idea, "The first time you draw this card each combat, draw 1 card" is closer to my original design for the Flight mod anyway.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
modargo: I'm still concerned that the fights will be too easy too often, but we can give it a try and see if it needs adjustment. Act 1 and 2 at least won't be too far off balance with low numbers.

Though could we at least make this zone not spawn in act 3? If there's no adjustment to the flight numbers to account for having an additional boss relic and a much stronger deck, this zone will be easier fights plus better rewards for a lot of decks.

Darkglade1: Having it not spawn in act 3 is fair enough, I suppose

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
modargo: Ignoring minions is a sensible way to do it. One thought that occurred to me is that I'd want the logic to ignore minions that are present at the start of combat too (e.g. for the Gremlin Leader fight). It would be a bit strange if the initially present minions got Flight, but newly spawned ones didn't. With that approach something like Gremlin Leader would count as a single-enemy fight too (since only one enemy is getting flight). That's what makes the most sense to me, at least.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
Darkglade1: Encounters with only a single non-minion enemy now get +1 Flight. The zone also no longer spawns in act 3.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/FlightMod.java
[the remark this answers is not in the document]
Darkglade1: Changed Flight mod to "The first time you draw this card each combat, draw 1 card"

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
Darkglade1: Byrds in the flight zone now get +1 strength and a line of trash talk. Minions are excluded from the flight mechanic.

The only other problematic enemy I noticed with my zone is Spheric Guardian and to a lesser extent Sentry due to their artifact allowing them to block the Stun from Flight's removal and Spheric Guardian's extremely low hp pool effectively negating the hp reduction of Flight. I don't really know of an elegant solution to this issue beyond hardcoding an exception for Spheric Guardian in particular.

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
Darkglade1: added this change so it works with preserved insect

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyFlightPower.java
Darkglade1: glad you like it 👍

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
```diff
@@ -36,17 +43,39 @@ public Color getColor() {
         return Color.SKY.cpy();
     }
 
+    @Override
+    public boolean canSpawn() {
+        // Don't spawn in Act 3
+        return AbstractDungeon.actNum < 3;
```
erasels, line 49: Hmm, maybe work with modulo so it doesn't stop spawning for endless?

modargo: There are convenient helper functions for this.
```suggestion
        return this.isAct(1) || this.isAct(2);
```
(I wrote the suggestion to check act 1 or 2 just in case anything weird happens with act numbers.)

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
modargo: Maybe have it also reduce the block enemies start with? Hopefully the timing works out so that you can catch Spheric Guardian's block that way. That isn't perfect since a lot of Spheric Guardian's "HP" is the block it gains during the fight, but it seems close enough.

For the Stun thing, it's weird to me that Stun is a debuff at all. When it's used for a mechanic like this, it shouldn't be. I'd support making your own version of `StunMonsterAction` that does something like:
```
StunMonsterPower power = new new StunMonsterPower(this.target, this.amount);
power.type = NeutralPowertypePatch.NEUTRAL;
AbstractDungeon.actionManager.addToTop(new ApplyPowerAction(this.target, this.source, power, this.amount));
```

Darkglade1: Updated to nerf spheric guardian's start of combat block and make stun go past artifact

src/main/java/spireMapOverhaul/zones/heavenlyClouds/HeavenlyCloudsZone.java
Darkglade1: changed to use this check

Cany0udance:
Getting this zone with Neow's Lament hardlocks the player. All monsters have their HP set to 0 but are still alive, meaning they cannot be targeted.

modargo:
> Getting this zone with Neow's Lament hardlocks the player. All monsters have their HP set to 0 but are still alive, meaning they cannot be targeted.

Funny bug. I've made a fix to ensure that enemies affected by this zone always have at least 1 HP (and 1 max HP, just in case someone makes a modded monster with only 1 max HP).

PR #70, opened 2024-01-08
Void Seed zone

tetraminus:
"This area has been corrupted by the void. It is not safe to rest here for long. 

 You will lose HP at the start of each turn. Enemies will be Voidtouched. Cards have a high chance of being Corrupted. The rest option is Corrupted."

Corrupted cards will double most numbers on the card, but they will cost 4 hp when played.
Voidtouched enemies gain 1 buffer at the start of their turn if they have none. (not the first turn)
The rest option is replaced with the option to corrupt another card.

Possible balancing:
I need a little input on this, but my ideas are to change the HP cost of corrupted cards, along with making the campfire corruption also cost HP.


src/main/java/spireMapOverhaul/zones/voidseed/campfire/CampfireCorruptEffect.java
```diff
@@ -0,0 +1,106 @@
...
+import com.megacrit.cardcrawl.rooms.RestRoom;
+import com.megacrit.cardcrawl.vfx.AbstractGameEffect;
+import com.megacrit.cardcrawl.vfx.cardManip.ShowCardBrieflyEffect;
+import spireMapOverhaul.zones.voidseed.cardmods.CorruptedModifier;
+
+import java.util.Iterator;
+
+import static spireMapOverhaul.SpireAnniversary6Mod.makeID;
+
+public class CampfireCorruptEffect extends AbstractGameEffect {
```
erasels, line 26: Please stop copying decompiled code from intellih's decompiler, it's killing me. At least copy it from jdgui or somethign so it isn't as obvious.

src/main/java/spireMapOverhaul/zones/voidseed/cardmods/CorruptedModifier.java
```diff
@@ -0,0 +1,119 @@
...
+        super.onUse(card, target, action);
+    }
+
+    @Override
+    public void onRender(AbstractCard card, SpriteBatch sb) {
+        VoidSeedShaderManager.StopFbo(sb, 0.0f, 1.0f, true);
+        super.onRender(card, sb);
+    }
+
+    @SpirePatches2({@SpirePatch2(
```
erasels, line 96: We doN#t use spirepatches manually, just stack the spirepatch annotations without containing class

erasels (changes requested):
I'll be honest, this zone doesn't seem very fun to me. The void touched modifier is bound to be ridiculous in many cases and unintuitive in others (many things in base game do not use magic number when they never change the value). The combat modifier seems somewhat boring as well and not very connected to the zone theme. 

I won't say this can't be part of the project or anything, but this is definitely a zone I wouldn't be happy to see in a run.

tetraminus:
Yeah this was the first one I came up with, I might just scrap it to be honest. I have a few other (honestly cooler) ideas that I want to try instead. Thanks for the feedback though.

src/main/java/spireMapOverhaul/SpireAnniversary6Mod.java
```diff
@@ -692,6 +696,13 @@ private void updateZoneList() {
     @Override
     public void receivePostUpdate() {
         time += Gdx.graphics.getRawDeltaTime();
+        if (AbstractDungeon.player != null) {
```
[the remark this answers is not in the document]
tetraminus: It caused a commodification crash. I think something with it looping over the rewards, activating the relic, and adding more to the rewards.

PR #73, opened 2024-01-10
Pack Attack zone

modargo:
I felt this project needed a Packmaster tribute zone, so I made one. I was originally planning to stop at two zones (and don't plan to make more after this), but having a Packmaster tribute was important to me.

This zone picks three packs, and card rewards and shop cards are taken from the zone's packs. In combat, you get a random card from those packs added to your hand each turn, in return for drawing one less card (I didn't want it to be purely positive). If this feels too limiting, it could be two random cards instead.

This remaining content is the zone is two events that are broadly positive and intended to draw people towards the zone. If you get an event while in the zone, it will always be one of these events.

Pack Shrine lets you either pick 1 out of 20 of any Packmaster cards (Library-style events always offer interesting choices) or remove a card by offering it to the shrine in return for gold based on the card's rarity (think Bonfire Spirits, except gold instead of healing and max HP). Like all shrines, this is a purely positive event. It has image variants for both Packmaster and Packmistress.

Black Market Two has three options: PM Booster Box in return for the Vexed curse, three random cards from the zone's packs in return for losing a small amount of HP, or Cardistry in return for some gold. These have costs, but are all intended to be good options. PM Booster Box is strong enough to be worth a curse, three random cards will often be good (but may depend on the zone's packs), and Cardistry is worth buying for most decks. The image is a flipped and slightly modified version of the Packmaster event.

Everything is achieved by reflection so that this doesn't require any dependencies. The zone can only spawn if Packmaster is loaded, and I've tested that it doesn't break when Packmaster isn't loaded. Packmaster's pack filtering is respected.

Most of the code here is for the events and getting pack/card information from Packmaster by reflection. The rest of the zone setup is straightforward.

Here are screenshots for the start of each event:
event 1

event 2


src/main/java/spireMapOverhaul/zones/packattack/PackAttackZone.java
```diff
@@ -0,0 +1,179 @@
...
+        this.replaceWithPackCards(cards, AbstractDungeon.cardRng);
+        this.applyStandardUpgradeLogic(cards);
+    }
+
+    @Override
+    public void postCreateShopCards(ArrayList<AbstractCard> coloredCards, ArrayList<AbstractCard> colorlessCards) {
+        this.replaceWithPackCards(coloredCards, AbstractDungeon.merchantRng);
+        for (AbstractCard c : coloredCards) {
+            for (AbstractRelic relic : AbstractDungeon.player.relics) {
+                relic.onPreviewObtainCard(c);
```
erasels, line 145: Why not use applyStandardUpgradeLogic here?

src/main/java/spireMapOverhaul/zones/packattack/PackAttackZone.java
```diff
@@ -0,0 +1,179 @@
...
+                options = this.commonPool;
+                break;
+            case UNCOMMON:
+                options = this.uncommonPool;
+                break;
+            case RARE:
+                options = this.rarePool;
+                break;
+            default:
+                throw new RuntimeException("Invalid card rarity for rewards: " + rarity);
```
erasels, line 174: I think by default we should just return a common card or something. Don't see a reason to make this an incompatibility

src/main/java/spireMapOverhaul/zones/packattack/events/BlackMarketTwo.java
```diff
@@ -0,0 +1,129 @@
...
+
+        if (!Loader.isModLoaded("anniv5")) {
+            throw new RuntimeException("This event requires Packmaster to be loaded");
+        }
+
+        this.gold = AbstractDungeon.ascensionLevel >= 15 ? A15_GOLD : GOLD;
+        this.hpLoss = AbstractDungeon.ascensionLevel >= 15 ? A15_HP_LOSS : HP_LOSS;
+        this.boosterBox = RelicLibrary.getRelic("anniv5:PMBoosterBox").makeCopy();
+        this.vexed = CardLibrary.getCard("anniv5:Vexed").makeCopy();
+        this.cardistry = CardLibrary.getCard("anniv5:Cardistry").makeCopy();
```
erasels, line 56: Using string literals for Ids, smh 😝

src/main/java/spireMapOverhaul/zones/packattack/events/PackShrine.java
```diff
@@ -0,0 +1,195 @@
...
+    private int screenNum = 0;
+
+    public PackShrine() {
+        super(NAME, DESCRIPTIONS[0], IMG_1);
+        if (MathUtils.randomBoolean()) {
+            this.imageEventText.loadImage(IMG_2);
+        }
+
+        if (!Loader.isModLoaded("anniv5")) {
+            throw new RuntimeException("This event requires Packmaster to be loaded");
```
erasels, line 54: Hmm, not sure if this is a good idea. If some mod does something weird with a dynamic way to get all events they might run into this accidentally.

Could we just set the only dialogue option to leave if that's the case instead?

erasels (changes requested):
Looks good, just figure it can be a bit less crash happy.

src/main/java/spireMapOverhaul/zones/packattack/PackAttackZone.java
modargo: This is for shop cards, which don't get the act-dependent update chance.

src/main/java/spireMapOverhaul/zones/packattack/PackAttackZone.java
modargo: That check is already in place, just that because of how I structured this it's in the calling method instead (`replaceWithPackCards`). Though looking at the code now, it's less code and more sensible for it to be in this method instead, so I'll do that. One of those things you don't always notice when writing the code.

src/main/java/spireMapOverhaul/zones/packattack/events/PackShrine.java
modargo: Good suggestion, that will be easy to do.

modargo:
Changes made. Leaving this for you to merge so you can take a quick look over the changes and confirm they match your expectations.

erasels:
Looks good, thanks!

PR #74, opened 2024-01-10
Beasts Lair

tetraminus:
A 1x1 zone that contains a boss from the previous act. it drops only uncommon and rare cards, along with 2 relics.


src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairZone.java
```diff
@@ -0,0 +1,142 @@
...
+        return 50;
+    }
+
+    @Override
+    public int changeUncommonCardRewardChance(int uncommonCardRewardChance) {
+        return 50;
+    }
+
+    @Override
+    public ArrayList<RewardItem> getAdditionalRewards() {
```
modargo, line 102: I like the theme for this zone, it's something that hasn't been done yet in this mod, and having a 1x1 zone is cool, so I'm on board with the overall concept. I do have some balance feedback, since I've actually made a fight exactly like this before (as part of an event for my act mods). What I discovered from extensive playtesting is that in most runs it's not very difficult to defeat a boss from the previous act.

Compared to when you'd normally face that boss, you have an extra boss relic plus whatever cards, relics, card removals, etc. you've gotten in the meantime. That makes a fight like this easier than an elite, given how bosses are designed (act 1 bosses are largely about achieving consistent damage and block, act 2 bosses are largely about scaling, aoe, and handling enemies with gimmicks). Think about act 2, for example. Which do you think is more likely to kill you -- Slavers or Guardian? In a typical run in act 2, the former is one of the most dangerous fights in the game, the latter is barely a threat.

That means right now, the risk/reward for this is off, since it's a fight that will typically be easier in return for greater rewards. For comparison, the event I mentioned gave the boss 2 additional strength and 25% more HP and the reward was a rare relic plus the normal card/gold rewards (but the rare relic is the only relic and no rarity adjustment to the cards). I felt that was reasonably balanced but still favorable for the player, meaning you wanted to take the fight every time unless your deck was really weak.

So this needs some tuning. The question is whether to do that by increasing the challenge, decreasing the rewards, or some of both. I think the flavor of this zone being a challenging encounter makes sense, so I support increasing the difficulty of this fight until it genuinely is a challenge. To justify the current rewards, though, you'd need to make this fight as challenging as the Colosseum event, since the current rewards are on a similar level (depending on how you value the uncommon/rare cards in the rewards vs. the extra gold from Colosseum).

That suggests one of two approaches:
* Increase the challenge of this fight moderately, so that it's as challenging as a normal elite or a bit more, and reduce the rewards to fit that. You could go with something similar to the 2 strength and 25% HP that I mentioned and tone down the rewards be a bit more than the normal amount for an elite.
* Increase the challenge of this fight immensely, so that it's as challenging as the Colosseum, and keep the current rewards. That means substantially more buffs than the 2 strength and 25% HP that I mentioned. I'm not sure how far you'd need to go, but I'd assume this would require at least 4 strength and maybe some other buffs too (like a bit of artifact or some metallicize). 

This can be adjusted in playtesting, so we don't need to get it exactly correct right now, but it's good to be in the right ballpark. That said, it's a real downside for the second option that it's harder to tune the challenge correctly.

Hope that all makes sense and happy to discuss further.

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairZone.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public void manualRoomPlacement(Random rng) {
+        for (MapRoomNode node : nodes) {
+            node.setRoom(new MonsterRoomElite());
+        }
+    }
+
+    @Override
+    public MonsterGroup changeEncounter(MonsterGroup monsterGroup) {
```
modargo, line 83: I think there will be a few oddities with this approach that need to be fixed. In particular, anything that checks the room type to determine what to do will think this is an elite fight, even though it really isn't. So for example, Black Star will give you an extra relic, even though you're fighting a boss, not an elite.

I'm sure there's a few different ways you can fix this issue.

One would be to make this an event (like Mind Bloom) rather than a fight. You can use `ForcedEventRoom` in this project (which I made for another zone) to do that. Since that follows a base game pattern, it's a good way to ensure that nothing too weird happens. You'd have to do a bit of work to make a little event for this, but that has some plusses too -- it means you get to write a bit of flavor about entering the Beast's Lair. This would also have the benefit of making this node look a bit different on the map (it will be a ? with little flames flickering behind it).

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairZone.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public Color getColor() {
+        return Color.RED.cpy();
+    }
+
+    @Override
+    public boolean canSpawn() {
+        return AbstractDungeon.actNum >= 2;
```
modargo, line 130: In endless mode this will mean this zone starts showing up in act 1 with act 3 bosses in it. If you want that, that's fine (endless mode is always a bit weird). If you don't want that, you can check `ActUtil.getRealActNum()` instead

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairZone.java
tetraminus: I think that would probably look better on the map too. I'll switch to that.

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairZone.java
tetraminus: I think that makes sense to me because it the previous act, but I think could go either way. ill leave it for now and change it if it feels weird.

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairZone.java
tetraminus: I want to make it feel like a real challenge, so I think I'll try and make it more colosseum level. I'll experiment with different buffs.

Darkglade1: The problem with a strength buff is that some bosses benefit significantly more from strength than others (imagine Hexaghost with an extra 4 strength on its divider or inferno, or guardian with 4 strength on its whirlwind). Meanwhile slime boss only has a single-hit attack, and loses the strength and other buffs as soon as it splits.

I think instead of strength it would be better to have a % damage increase buff (like this creature deals 33% more damage). Also any new enemies that spawn over the course of the fight should also obtain the buffs.

tetraminus: That's true, I'll try that. honestly I might even try things that aren't even straight damage increases

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairEvent.java
```diff
@@ -0,0 +1,78 @@
...
+        AbstractDungeon.actionManager.addToBottom(new AllEnemyApplyPowerAction(null, 10, (m) -> new FuryPower(m, 10)));
+        Wiz.forAllMonstersLiving((m)->{
+            m.maxHealth = (int) (m.maxHealth * 1.75f);
+            m.currentHealth = (int) (m.currentHealth * 1.75f);
+        }
+        );
+    }
+
+    @SpirePatch2(clz = SpawnMonsterAction.class, method = "update")
+    public static class SpawnMonsterActionPatch {
```
modargo, line 57: What does this do to the slimes spawned by Slime Boss? Since their HP is based on Slime Boss's HP and Slime Boss will already have 75% more HP, you wouldn't want the spawned slimes to also get extra HP.

You may need to special case that by checking if you're in the Slime Boss fight (which you can probably detect through checking `AbstractDungeon.lastCombatMetricKey.equals(SlimeBoss.ID)`) and the enemies being spawned are any of the relevant slimes (the large/medium acid/spike slimes).

src/main/resources/anniv6Resources/localization/eng/BeastsLair/Powerstrings.json
```diff
@@ -0,0 +1,15 @@
+{
+
+  "${ModID}:FuryPower": {
+    "NAME": "Fury",
+    "DESCRIPTIONS": [
+      "Whenever you play a card, gains #b", "% Damage up until end of turn."
```
modargo, line 6: I'd word this the same way as the Damage Up power. I think that's clearer than saying that it gains Damage Up
```suggestion
      "Whenever you play a card, deals #b", "% more damage until end of turn."
```

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairEvent.java
```diff
@@ -0,0 +1,78 @@
...
+import com.megacrit.cardcrawl.localization.EventStrings;
+import com.megacrit.cardcrawl.monsters.AbstractMonster;
+import com.megacrit.cardcrawl.powers.ArtifactPower;
+import javassist.CtBehavior;
+import spireMapOverhaul.actions.AllEnemyApplyPowerAction;
+import spireMapOverhaul.util.Wiz;
+import spireMapOverhaul.zones.beastslair.powers.FuryPower;
+
+import static spireMapOverhaul.SpireAnniversary6Mod.makeID;
+@AutoAdd.Ignore
```
modargo, line 22: I don't think you need this. When the event is part of a zone, it won't show up unless you specifically set it to do so with the `ModifiedEventRateZone` interface or a forced event like this zone uses.

The issue with the example event was because it wasn't part of a zone anymore, so it got registered as a generic event that could show up anywhere.

src/main/java/spireMapOverhaul/zones/beastslair/BeastsLairEvent.java
```diff
@@ -0,0 +1,78 @@
...
+public class BeastsLairEvent extends PhasedEvent {
+        public static final String ID = makeID("BeastsLairEvent");
+        //These eventStrings should be defined in a json file and loaded in your main mod file. See https://github.com/daviscook477/BaseMod/wiki/Custom-Strings
```
modargo, line 25: You can get rid of comments that came from the template like this.

modargo:
Looks good to me. Thanks for taking care of all the feedback.

The buff you settled on for the boss fight is interesting. It certainly has the potential to be deadly, though it will affect some decks more than others (pity the poor shiv deck). But as a unique way to change the fight that will make people think about whether or not to risk this, it seems like a good way to go.

Depending on playtesting we may find that it needs balance adjustment (and I find it tricky to predict how deadly it makes this encounter on average), but hopefully it's in the right ballpark.

PR #80, opened 2024-01-12
Ambushed zone

Cany0udance:
First time ever doing a PR. Hopefully I didn't botch anything.

Known points of concern for the zone:

- Sentry + Lagavulin may be too difficult for act 1, especially if the map layout forces the player into it
- Occasionally, the shop may disrespect traditional shop rules. Instead of attack, attack, skill, skill, (power which my zone replaces with attack/skill) in the top row, it may become attack, skill, skill, skill, (other card) or skill, attack, skill, skill, (other card)
- If the player has The Courier, newly generated cards come from the player's class as normal and not card draw cards as you would expect from my zone. Similar issue with potions


src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -0,0 +1,384 @@
...
+
+
+    private boolean isCardForShop(String cardID, AbstractCard.CardRarity rarity, AbstractCard.CardType type, boolean isColored) {
+        AbstractCard card = CardLibrary.getCard(cardID);
+        boolean isCardColored = card.color != AbstractCard.CardColor.COLORLESS;
+        return card != null && card.rarity == rarity && card.type == type && isCardColored == isColored;
+    }
+
+    @Override
+    public void postCreateShopPotions(ShopScreen screen, ArrayList<StorePotion> potions) {
```
erasels, line 233: I think this is not a good idea. Maybe just replace one potion with a draw potion. This seems overkill, and yoU're unlikely to want this many anyway. Also too much repetition in the potion choices

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -0,0 +1,384 @@
...
+                    break;
+                default:
+                    continue; // Skip if the potion has an unexpected rarity
+            }
+            storePotion.potion = replacementPotion;
+        }
+    }
+
+    @Override
+    public void modifyReward(RewardItem rewardItem) {
```
erasels, line 259: Again, don't think this is a good idea since it's always the same potions.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -0,0 +1,384 @@
...
+        );
+    }
+
+    @Override
+    public List<ZoneEncounter> getEliteEncounters() {
+        return Arrays.asList(
+                new ZoneEncounter(AMBUSHED_BY_SENTRIES, 1, () -> new MonsterGroup(
+                        new AbstractMonster[] {
+                                new Sentry(-1000.0F, 0),
+                                new Lagavulin(true)
```
erasels, line 365: Maybe use the combatmodifying zone to check for the laga encounters and nerf it a bit? This does seem a lot harder than the base game elites

erasels (changes requested):
Cool zone!

I'm unsure about the reward modification as a whole here though. Seeing only base game draw cards in the rewards seems very one note and only seeing draw potions aso doesn't feel great.

What about adding a single draw card as an additional card reward that players may pick in addition to the normal card reward?
https://github.com/erasels/SpireMapOverhaul/blob/main/src/main/java/spireMapOverhaul/rewards/SingleCardReward.java

That may also help justify the harder than normal fights.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    @Override
+    protected boolean canIncludeEarlyRows() {
+        //Since this zone always has an elite, we don't want it to show up in the rows of the act that never have elites
+        //It also has normal fights that are tuned as hard pool fights, so that's another reason to prevent early spawns
+        return false;
+    }
+
+    private void initCardDrawCardIDs() {
```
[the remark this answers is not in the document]
Cany0udance: There are no powers in the game that draw cards _immediately_. The rewards are meant to be tools that can be used in a pinch to grant an edge in tough situations. Cards like Machine Learning do nothing immediately and are too slow to be fitting for the zone, in my opinion.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
Cany0udance: I would make the argument that due to the amount of zones there are and how small my zone is in terms of height + width, the odds of someone naturally wandering into a shop that happens to fall in my zone are very slim.

Also, replacing just one potion with a draw potion would be too petty to add as a modification I think. I am satisfied with the current setup, although maybe I could make it so one potion of each rarity is guaranteed to appear? So the player has the option to purchase a Swift Potion, a Gambler's Brew, and a Smoke Bomb (lol) / Snecko Oil. I do think three Swift Potions in one shop feels pretty bad.

Cany0udance:
So for the reward modification, here's how I see it.

The player is really looking for those high roll card draw cards that will help them throughout the rest of the game. We're talking Master of Strategy, Scrawl, Reboot, Acrobatics even. Therefore, I believe having 3 options per fight vastly improves how "good" the zone feels, as the odds of getting a high roll draw card are fairly reasonable. 

If my zone only had 1 card draw choice per fight, you could see something like Prepared, then Impatience, then Escape Plan, then oops, you're out of the zone and you took all those hard combats for basically nothing. I think the current configuration is reasonable. As the zone only appears in acts 1 and 2, the zone gives you the chance to pick cards that will help your deck throughout the rest of the run, as card draw is always nice to have.

Also to answer all of your "why did you code it like this" questions, I do not know Java myself so I had ChatGPT program this whole thing. I will do my best to address some of your concerns and then update the PR.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
Cany0udance: I don't like the idea of having inconsistency in the game, where loading into a specific encounter nerfs a specific enemy. I'd rather try to find another enemy to replace Sentry with. I don't think the Sentry itself is that bad of an addition, it's the Dazes it adds that will REALLY mess a player up, especially if Laga debuffs them and their next turn is just full of Dazes.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
erasels: Yeah, I'd be okay with that.

erasels:
I see your point in regards to trying to get the highroll draw cards, however I believe that there's an issue with it. The thing you may be discounting here is that drawing cards is only good if you have things worth drawing. This zone can spawn in both act 1 and 2 where players are likely to be still building their deck and pick up value cards that help them win. You would not want to pick up an acrobatics when you have few high value cards or syngeristic effects.
While I agree that it's likely fine in act 2, in act 1 I find it hard to justify going into a hard zone to get draw cards when you don't even have a real deck to draw yet. That's a lot more risk than reward for a payoff that might require another act to payoff because you "skipped" ~3 combat prowess increasing card rewards.

Also, your code quality is fine, that locator just struck me as very weird. The others are just small fixes, I'm sure we'll get there.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -0,0 +1,385 @@
...
+                case COMMON:
+                    replacementPotion = PotionHelper.getPotion(SwiftPotion.POTION_ID);
+                    break;
+                case UNCOMMON:
+                    replacementPotion = PotionHelper.getPotion(GamblersBrew.POTION_ID);
+                    break;
+                case RARE:
+                    // Randomly select between Smoke Bomb and Snecko Oil for Rare potions
+                    if (AbstractDungeon.cardRandomRng.randomBoolean()) {
+                        replacementPotion = PotionHelper.getPotion(SmokeBomb.POTION_ID);
```
modargo, line 279: Getting smoke bombs feels like a troll. Taking harder fights and having part of the "reward" be a near-useless potion isn't a great experience.

I'm not sure about the whole potion replacement thing either, but if you're doing it, it would land better if you cut the smoke bomb thing and just gave out Snecko Oils.

Cany0udance:
I made the following changes:

- Sentry + Lagavulin is now Fungi Beast + Lagavulin. I still feel like this is slightly too difficult. I don't want to add a Louse because it would it's insignificant in the fight as a whole and the Nob encounter already has a Louse. I don't want to add a Slime either because 3 of my 5 normal act 1 encounters include a Slime which is already too much probably.
- Red + Blue Slavers is now Byrd + Book of Stabbing. Again, I feel like this is slightly too difficult.
- String literals nuked
- Smoke Bombs removed from the zone entirely. I had them there for thematic purposes but losing them isn't that big of a deal
- Shops will now always offer a Swift Potion, Gambler's Brew, and Snecko Oil. Though notably, the base price of these is ALWAYS the same and isn't picked within a range like regular shop potions are. Not sure how to fix that.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -0,0 +1,385 @@
...
+                new ZoneEncounter(AMBUSHED_BY_CULTIST_CHOSEN, 2, () -> new MonsterGroup(
+                        new AbstractMonster[] {
+                                new Cultist(-1000.0F, 0),
+                                new Chosen(32.0F, 0)
+                        }))
+        );
+    }
+
+    @Override
+    public List<ZoneEncounter> getEliteEncounters() {
```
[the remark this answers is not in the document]
modargo: The updated elite encounters make sense to me.

Having thought about it more, keeping Taskmaster and Nob seems fine. It can be a dangerous fight (especially if you aren't able to kill Nob quickly), but being dangerous is in line with the other elite encounters in this zone, and I don't think it's out of line with the other elite ambushes you've made. So let's give it a shot, and if it turns out to be too much, we can replace or change it.

One final thought on these encounters. Given that they're harder than normal, it may be worth considering at least a small additional reward. I get that if you're looking for card draw, getting a card reward with multiple card draw options is already a boost. But it might be worth something more on top of that, such as:
* A small amount of extra gold
* Having card rewards be more likely to be upgraded
* An extra guaranteed potion

I don't think it's critical to the zone to have this, but right now my intuition is that the risk/reward of this zone may be slightly too skewed towards risk. Can't be sure of that without playing with it a bunch, but together with Gk's feedback, it supports the idea that a bit more reward may help balance things out. (Depending on what you go with, having normal encounters also get a small extra reward might make sense.)

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -0,0 +1,380 @@
...
+            // Add Swift Potion with its appropriate price
+            potions.add(new StorePotion(PotionHelper.getPotion(SwiftPotion.POTION_ID), 0, screen));
+
+            // Add Gambler's Brew with its appropriate price
+            potions.add(new StorePotion(PotionHelper.getPotion(GamblersBrew.POTION_ID), 1, screen));
+
+            // Add Snecko Oil with its appropriate price
+            potions.add(new StorePotion(PotionHelper.getPotion(SneckoOil.POTION_ID), 2, screen));
+        }
+    }
```
modargo, line 251: Try adding the following to get the normal price variation for these potions:
```

        for (StorePotion potion: potions) {
            if (!Settings.isDailyRun) {
                potion.price = MathUtils.round(potion.price * AbstractDungeon.merchantRng.random(0.95F, 1.05F));
            }
        }
```

erasels: I did this myself

PR #81, opened 2024-01-12
Prevent zones from showing up more than once in a run

modargo:
I took a bit of time to implement this since even if we end up not wanting it in the full release, I think it's a good thing to have in place for playtesting. (I certainly want it for my own playtesting, at the very least.)

We can discuss exactly how we want to merge this -- as is, or with some kind of setting or config option or console command for it, or some other option. I do want to get some version of these changes merged in, though, since it has a better system for recording which zones are valid for your current run (using a savable instead of a config).

### DIFF
--- a/src/main/java/spireMapOverhaul/BetterMapGenerator.java
+++ b/src/main/java/spireMapOverhaul/BetterMapGenerator.java
@@ -11,6 +11,7 @@ import spireMapOverhaul.patches.ZonePatches;
 
 import java.util.*;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 public class BetterMapGenerator {
     public static final Logger mapGenLogger = LogManager.getLogger("BetterMapGen");
@@ -36,12 +37,16 @@ public class BetterMapGenerator {
     }
 
     public ArrayList<ArrayList<MapRoomNode>> generate(Random rng, int width, int height, int pathDensity) {
+        SpireAnniversary6Mod.currentRunSeenZones.addAll(activeZones.stream().map(z -> z.id).collect(Collectors.toList()));
+        mapGenLogger.info("Already seen zones: " + String.join(",", SpireAnniversary6Mod.currentRunSeenZones));
         MapPlanner planner;
         do {
             List<AbstractZone> possibleZones = new ArrayList<>();
-            for (AbstractZone zone : SpireAnniversary6Mod.allZones)
-                if (zone.canSpawn())
+            for (AbstractZone zone : SpireAnniversary6Mod.unfilteredAllZones) {
+                if (SpireAnniversary6Mod.currentRunAllZones.contains(zone.id) && !SpireAnniversary6Mod.currentRunSeenZones.contains(zone.id) && zone.canSpawn()) {
                     possibleZones.add(zone);
+                }
+            }
 
             planner = new MapPlanner(width, height);
 
--- a/src/main/java/spireMapOverhaul/BetterMapGenerator.java
+++ b/src/main/java/spireMapOverhaul/BetterMapGenerator.java
@@ -51,7 +51,7 @@ public class BetterMapGenerator {
         do {
             List<AbstractZone> possibleZones = new ArrayList<>();
             for (AbstractZone zone : SpireAnniversary6Mod.unfilteredAllZones) {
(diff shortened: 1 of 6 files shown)


src/main/java/spireMapOverhaul/util/ActUtil.java
```diff
@@ -1,11 +1,40 @@
...
+import com.evacipated.cardcrawl.modthespire.Loader;
 import com.megacrit.cardcrawl.core.Settings;
-import com.megacrit.cardcrawl.dungeons.AbstractDungeon;
+import com.megacrit.cardcrawl.dungeons.*;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 
 public class ActUtil {
     public static int getRealActNum() {
```
modargo, line 11: This implementation is adapted from what I did for Corrupt the Spire. I found it to be a reliable way of getting the real act number -- haven't had any errors related to it since I implemented it.

src/main/java/spireMapOverhaul/BetterMapGenerator.java
```diff
@@ -36,12 +39,22 @@ private BetterMapGenerator() {
     }
 
     public ArrayList<ArrayList<MapRoomNode>> generate(Random rng, int width, int height, int pathDensity) {
+        if (Settings.isEndless && ActUtil.getRealActNum() == 1) {
```
modargo, line 42: Properly implementing this check -- making endless mode able to see repeat zones -- involved a more robust version of `getRealActNum` that you can see in `ActUtil`. This seemed like a sensible way to do it, since I think endless mode lets you see repeat events too.

There will probably be some janky aspects to doing certain zones multiple times, but endless is always a bit weird, so I figure we fix those as they come up. At least, that seems like it will result in fewer complaints than having endless get no zones at all after seeing each zone once.

src/main/java/spireMapOverhaul/util/ActUtil.java
```diff
@@ -1,11 +1,40 @@
...
+                    return 1;
+                case TheCity.ID:
+                case "theJungle:Jungle":
+                    return 2;
+                case TheBeyond.ID:
+                    return 3;
+                case TheEnding.ID:
+                    return 4;
+                default:
+                    throw new RuntimeException("Unrecognized act ID: " + AbstractDungeon.id + ". When ActLikeIt isn't loaded, the only possible acts should be the vanilla four and The Jungle.");
```
erasels, line 36: I wonder if this will crash with VUP Shion and Animator's special acts

erasels:
I think I'd like this to be a mod config option which can default to on. No dupes in a run is probably for the best but depending on how many zones we ship with, this might to default to off.

src/main/java/spireMapOverhaul/BetterMapGenerator.java
```diff
@@ -36,12 +39,22 @@ private BetterMapGenerator() {
...
+            SpireAnniversary6Mod.currentRunSeenZones.addAll(activeZones.stream().map(z -> z.id).collect(Collectors.toList()));
+        }
+        mapGenLogger.info("Already seen zones: " + String.join(",", SpireAnniversary6Mod.currentRunSeenZones));
         MapPlanner planner;
         do {
             List<AbstractZone> possibleZones = new ArrayList<>();
-            for (AbstractZone zone : SpireAnniversary6Mod.allZones)
-                if (zone.canSpawn())
+            for (AbstractZone zone : SpireAnniversary6Mod.unfilteredAllZones) {
+                if (SpireAnniversary6Mod.currentRunAllZones.contains(zone.id) && !SpireAnniversary6Mod.currentRunSeenZones.contains(zone.id) && zone.canSpawn()) {
```
erasels, line 54: Might be good to give people a utility method that gives you all zones that are not filtered for the current run, so that people don't have to do this logic themselves

src/main/java/spireMapOverhaul/util/ActUtil.java
modargo: I'll have to test that. I'd believe it for Animator, everything in that mod is custom.

I looked through the VUP Shion code and I don't see any extra acts. It has a lot of logic for custom bosses and stuff, but no acts that I can find.

modargo:
> I think I'd like this to be a mod config option which can default to on. No dupes in a run is probably for the best but depending on how many zones we ship with, this might to default to off.

I'll add in the config option. We will have to see how many zones we end up with.

modargo:
Should be in good shape now. Config option added, a bit more testing done, some bugs fixed, and a bit more robustness.

PR #83, opened 2024-01-13
Fix forced events not being added to seenEvents

erasels:
It seems that forced events don't get added to SeenEvents.seenEvents spirefield on AbstractPlayer. It happens due to getEvent never being called if we force an event.

To fix this, I manually add it in the `modifyGen` patch, however getting the correct event ID was hard in that place so I rewrote some code so that we're working with IDs for the most part.


erasels:
Alch looks over it and couldn#t find an issue, so I'll be merging it.

PR #86, opened 2024-01-14
Add The Fog zone

ActuallyEnbeon:
In this zone:
- All nodes are replaced with ? nodes
- Enemies start combat with 1 Weak and 1 Vuln
- An exclusive event, "Standing Stones", can be encountered

This PR also contains a small change to AbstractSMORelic, that being a new constructor which takes in a zoneID String to allow relics' image files to be separated depending on their zone of origin.


ActuallyEnbeon:
Wow I made a lot of commits haha

src/main/java/spireMapOverhaul/zones/thefog/events/StandingStones.java
```diff
@@ -0,0 +1,88 @@
...
+            this.healAmt = MathUtils.round(adp().maxHealth * A15_HP_HEAL_PERCENT);
+            this.damageAmt = MathUtils.round(adp().maxHealth * A15_HP_DAMAGE_PERCENT);
+        } else {
+            this.healAmt = MathUtils.round(adp().maxHealth * HP_HEAL_PERCENT);
+            this.damageAmt = MathUtils.round(adp().maxHealth * HP_DAMAGE_PERCENT);
+        }
+
+        this.imageEventText.setDialogOption(OPTIONS[0] + this.healAmt + OPTIONS[1]);
+        this.imageEventText.setDialogOption(OPTIONS[2]);
+        this.imageEventText.setDialogOption(OPTIONS[3] + this.damageAmt + OPTIONS[4]);
```
modargo, line 50: Generally when the player is getting a specific fixed relic, the event will preview it on hover and sometimes say what it is in the text. See e.g. Golden Idol. For a modded event, we should at least have the preview on hover; if this isn't transparent about what the player is getting, people are less likely to engage with the content. So let's do that here.

You can provide a copy of the relic as the second parameter to `setDialogOption` to get the preview on hover. Up to you if you want to include the relic name in the text for the option.

src/main/resources/anniv6Resources/localization/eng/TheFog/Eventstrings.json
```diff
@@ -0,0 +1,17 @@
...
+      "You take the opportunity to get some #gwell-earned #grest. NL Settling down on the rough ground, you look up at the #pstarry #psky as you drift off. NL NL When you awake, the sky is #bobscured #bby #bthe #bfog. NL Time to get going.",
+      "Looking up at the sky, you can pick out familiar constellations, as well as #b~strange~ #bnew #bones. NL You wonder what #psecrets this place holds. NL NL Eventually, the stars begin to #rdim. NL Perhaps you've learned enough.",
+      "These stones hold #ppower, you're sure of it. NL You try to loosen a section of rock from one of the stones and eventually get it free - at the cost of a #rfew #rbruises. NL NL You pick up the fragment of stone and move on, its #baura a comfort on your journey through this place."
+    ],
+    "OPTIONS": [
+      "[Sleep] #gHeal #g", " #gHP.",
+      "[Stargaze] #gChoose #g1 #gof #g3 #gcards #gof #gany #gcolor #gto #gobtain.",
```
modargo, line 12: This wording is a bit weird. The base game only uses "obtain" for relics (cards are generally added to your deck, not obtained), and if this is a card reward, you don't want to say 1 of 3 because it could be a different number due to Question Card/Busted Crown (the Library says choose 1 of 20 because that's not a card reward, it's a special screen). 

Let's use something patterned off Sensory Stone's wording:
```suggestion
      "[Stargaze] #gAdd #g1 #gcard #gof #gany #gcolor #gto #gyour #gdeck.",
```

src/main/resources/anniv6Resources/localization/eng/UIstrings.json
```diff
@@ -20,6 +20,12 @@
       "Heal "
     ]
   },
+  "${ModID}:AnyColorCardReward": {
+    "TEXT": [
+      "Add a card of any color to deck",
```
modargo, line 25:
```suggestion
      "Add a card of any color to your deck",
```

src/main/java/spireMapOverhaul/zones/thefog/events/StandingStones.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final float A15_HP_DAMAGE_PERCENT = 0.2F;
+
+    private final int healAmt;
+    private final int damageAmt;
+
+    public StandingStones() {
+        super(NAME, DESCRIPTIONS[0], makeImagePath("events/TheFog/StandingStones.png"));
+        this.noCardsInRewards = true;
+
+        if (AbstractDungeon.ascensionLevel >= 15) {
```
modargo, line 40: Let's round all this down instead.

The base game rounds damage down in both fights and events, and while the healing bonus from Magic Flower is rounded up in fights, healing in events is mostly rounded down. Library uses `MathUtils.round` like this, but contrast that with Big Fish, Cleric, and Mushrooms, which all round down.

So while there's an argument you could make for rounding up the healing, I think the sanest thing to do is round everything down. That's most consistent with other events for both the damage and the healing, and it keeps this event internally consistent (I wouldn't want to have multiple rounding approaches in the same event).

Really just need to replace the round calls with `(int)`, e.g.:
```
this.healAmt = (int)(adp().maxHealth * A15_HP_HEAL_PERCENT);
this.damageAmt = (int)(adp().maxHealth * A15_HP_DAMAGE_PERCENT);
```
(and so on for the else clause)

modargo:
Looks like a fun zone but I wonder about the balance. The fights are purely easier for the player, nodes all being events is roughly neutral, the event is on par with other helpful events, and there's nothing negative for the player. 

Did you consider giving the player some minor debuff to compensate for this? It could be really small, like shuffling 1 Dazed into your draw pile at the start of combat. Something like that would round out the zone and make it not just easier (even if it will usually be a favorable trade-off). You could come up with some other minor drawback for the player but I actually like the Dazed idea because it seems very fitting with the fog theme.

src/main/java/spireMapOverhaul/zones/thefog/events/StandingStones.java
```diff
@@ -0,0 +1,88 @@
...
+import com.megacrit.cardcrawl.localization.EventStrings;
+import com.megacrit.cardcrawl.rooms.AbstractRoom;
+import spireMapOverhaul.rewards.AnyColorCardReward;
+import spireMapOverhaul.zones.thefog.relics.StoneFragment;
+
+import static spireMapOverhaul.SpireAnniversary6Mod.makeID;
+import static spireMapOverhaul.SpireAnniversary6Mod.makeImagePath;
+import static spireMapOverhaul.util.Wiz.adp;
+
+public class StandingStones extends AbstractImageEvent {
```
modargo, line 17: Am I right that you haven't done anything to make this event show up more often? In other words, it will just be in the pool of events while you're in the zone, no more or less likely to show up than any of the other events for the act the player is in.

Since the whole zone will be ? nodes and the player is likely to encounter some events in the zone, that's fine if you want it to work like that. On the other hand, if you want this to show up a bit more often, you can add something like this to your zone class:
```

    @Override
    public float zoneSpecificEventRate() {
        return 0.3f;
    }
```
That would give a 30% chance per event that you get a zone-specific event; I picked 30% as something that seemed like a reasonable bump to the normal probability without guaranteeing you see this event every time you're in the zone.

Up to you though, I'm making this comment just to call out the option to do this.

erasels:
> Looks like a fun zone but I wonder about the balance. The fights are purely easier for the player, nodes all being events is roughly neutral, the event is on par with other helpful events, and there's nothing negative for the player.

Nodes aren't all events, they look like event nodes, as in, you don't know what node it actually is. That's the downside and I think it's fine for the debuff it gives the enemies.

modargo:
> > Looks like a fun zone but I wonder about the balance. The fights are purely easier for the player, nodes all being events is roughly neutral, the event is on par with other helpful events, and there's nothing negative for the player.
> 
> Nodes aren't all events, they look like event nodes, as in, you don't know what node it actually is. That's the downside and I think it's fine for the debuff it gives the enemies.
> 
> Or wait, I'm mistaken? I guess the zone idea has changed since its conception.

You're right, but that was my understanding all along. I can see how what I said was ambiguous, but by "all the nodes are events" I meant EventRooms aka ? rooms aka "you have a chance to get an event, but could also get a shop, fight, or chest." I don't see that as a downside. When I see a stretch of ? nodes during a run, my evaluation of it is roughly neutral. You don't know for sure what you're getting, but you know the probabilities and the overall distribution is certainly positive for the player.

Sometimes my evaluation of a stretch of ? nodes will be higher if I have things that make events better (e.g. Omamori or a deck that would benefit from Bites/Apparitions in act 2) or if I want to avoid fights (e.g. early act 2 when I just took Slaver's Collar and elite fights will be easier than hard pool fights). Sometimes my evaluation will be lower if I'm looking for specific non-event nodes (e.g. if I want a guaranteed shop, or if I want campfires to upgrade critical cards, or if I'm aiming for elites and need to take a different path to get them).

Of course there's a billion other factors if you really want to go deep on analyzing this (e.g., entering a stretch of ? nodes with no gold is generally unwise, since you might get a useless shop). But the bottom line is that I think a zone being ?s is pretty close to neutral. The whole map being ?s (like that one modded relic does) is an issue because it screws up all sorts of strategic pathing decisions, but that's completely different from just having one 3x4 section of the map as ?s (that's not too different from the normal variance you see in map layouts).

src/main/resources/anniv6Resources/localization/eng/UIstrings.json
ActuallyEnbeon: Although the current version looks like an oversight, it was actually a conscious decision on my part due to space restraints - see the suggested version in-game, the k doesn't quite fit in the reward box:
image

ActuallyEnbeon:
> Nodes aren't all events, they look like event nodes, as in, you don't know what node it actually is. That's the downside and I think it's fine for the debuff it gives the enemies.
> 
> Or wait, I'm mistaken? I guess the zone idea has changed since its conception.

One thing I did notice during testing is that it feels like there are _more_ events that I would expect - it makes sense, given that they're "Event nodes", but I have found it's possible to cross the entire zone without encountering a single combat if you're lucky. I couldn't find any hooks to modify the chance of a ? node giving an event though (aside from the zone-specific ones in ModifiedEventRateZone) so I left it as-is.

> Did you consider giving the player some minor debuff to compensate for this? It could be really small, like shuffling 1 Dazed into your draw pile at the start of combat. Something like that would round out the zone and make it not just easier (even if it will usually be a favorable trade-off). You could come up with some other minor drawback for the player but I actually like the Dazed idea because it seems very fitting with the fog theme.

Not a bad idea, I like the flavour of it! I probably will make that change, combats in the zone do seem a little bit too easy, from my experience during testing.

modargo:
> > Nodes aren't all events, they look like event nodes, as in, you don't know what node it actually is. That's the downside and I think it's fine for the debuff it gives the enemies.
> > Or wait, I'm mistaken? I guess the zone idea has changed since its conception.
> 
> One thing I did notice during testing is that it feels like there are _more_ events that I would expect - it makes sense, given that they're "Event nodes", but I have found it's possible to cross the entire zone without encountering a single combat if you're lucky. I couldn't find any hooks to modify the chance of a ? node giving an event though (aside from the zone-specific ones in ModifiedEventRateZone) so I left it as-is.

Yeah, we haven't made any kind of hook for that. It wouldn't be too hard to do, but IMO it's fine to just let the normal ? node logic happen. That fits what players will expect and makes it possible to accurately evaluate how this zone will play based on experience from the base game.

With regards to your testing, worth noting that the chance of getting non-event outcomes rises each time you get any other outcome. In the average run, a player encountering this zone will already have probabilities for some of the non-event outcomes boosted due to previous ? nodes. In contrast, if your testing involved going straight to this zone (without going through other ? nodes), that means your event probability will be higher than what people normally experience.

src/main/resources/anniv6Resources/localization/eng/TheFog/Eventstrings.json
```diff
@@ -0,0 +1,17 @@
+{
+  "${ModID}:StandingStones": {
+    "NAME": "Standing Stones",
+    "DESCRIPTIONS": [
+      "An #podd #pformation #pof #pstones appears from within the fog. Whilst standing in the centre of the circle they form, you can see the #bnight #bsky above - the ~real~ night sky, from before the #rSpireblight. NL NL Though you don't understand this place in the slightest, you feel #gsafe here. Why not stay for a while to recuperate?",
```
modargo, line 5: Just switching this to American English for consistency.
```suggestion
      "An #podd #pformation #pof #pstones appears from within the fog. Whilst standing in the center of the circle they form, you can see the #bnight #bsky above - the ~real~ night sky, from before the #rSpireblight. NL NL Though you don't understand this place in the slightest, you feel #gsafe here. Why not stay for a while to recuperate?",
```

modargo:
Looking good. Going to do a quick final pass then merge if I don't see anything else.

PR #92, opened 2024-01-15
Rework BeastsLair fury mechanic.

tetraminus:
Bosses now gain 10% damage up at the end of their turn, instead of when the player plays a card. this should lessen the problem of the original effect punishing high card builds a little too much.

### DIFF
--- a/src/main/resources/anniv6Resources/localization/eng/BeastsLair/Powerstrings.json
+++ b/src/main/resources/anniv6Resources/localization/eng/BeastsLair/Powerstrings.json
@@ -3,13 +3,13 @@
   "${ModID}:FuryPower": {
     "NAME": "Fury",
     "DESCRIPTIONS": [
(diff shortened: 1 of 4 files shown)


src/main/resources/anniv6Resources/localization/eng/BeastsLair/Powerstrings.json
```diff
@@ -3,13 +3,13 @@
   "${ModID}:FuryPower": {
     "NAME": "Fury",
     "DESCRIPTIONS": [
-      "Whenever you play a card, deals #b", "% more damage until end of turn."
+      "At the end of it's turn, gains #b", "% damage up for the rest of battle."
```
erasels, line 6:
```suggestion
      "At the end of its turn, gains #b", "% damage up for the rest of combat."
```

erasels (changes requested):

PR #93, opened 2024-01-15
Change (nerf) Stone Fragment

ActuallyEnbeon:
The special event Relic from The Fog zone

### DIFF
--- a/src/main/java/spireMapOverhaul/zones/thefog/relics/StoneFragment.java
+++ b/src/main/java/spireMapOverhaul/zones/thefog/relics/StoneFragment.java
@@ -15,7 +15,7 @@ public class StoneFragment extends AbstractSMORelic {
 
     public StoneFragment() {
         super(ID, TheFogZone.ID, RelicTier.SPECIAL, LandingSound.HEAVY);
-        counter = 13;
+        setCounter(13);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/spireMapOverhaul/zones/thefog/relics/StoneFragment.java
```diff
@@ -13,30 +15,33 @@ public class StoneFragment extends AbstractSMORelic {
...
-        counter += 1;
-        if (counter == 6) {
-            flash();
-            pulse = false;
-            stopPulse();
-            atb(new GainEnergyAction(1));
-        }
+    public void update() {
+        super.update();
+        description = (counter < 1) ? MSG[2] : DESCRIPTIONS[1]; // MSG[2] is the "used up" string
```
erasels, line 37: I guess you do this so it works correctly on save and load? Maybe we can find an alternative like checking justEnteredRoom + onVictory or something. I generally like to avoid doing something every frame if it's not needed.

src/main/resources/anniv6Resources/localization/eng/TheFog/Relicstrings.json
```diff
@@ -3,7 +3,8 @@
     "NAME": "Standing Stone Fragment",
     "FLAVOR": "The fragment exudes a powerful aura, but it seems just out of reach.",
     "DESCRIPTIONS": [
-      "When you draw your sixth card each turn, gain [E] [REMOVE_SPACE]."
+      "Starts with 13 charges. ",
```
erasels, line 6:
```suggestion
      "Starts with #b13 charges. ",
```

erasels (changes requested):

src/main/java/spireMapOverhaul/zones/thefog/relics/StoneFragment.java
ActuallyEnbeon: Just committed some replacement code - thoughts?

PR #96, opened 2024-01-15
Add Gremlin Camp Zone

erasels:
Changed EncounterModifyingZone:changeEncounter to also pass the id of the encounter, this allows for easier control in case you want to change specific encounters you add.

Also adds ModifiedEventRateZone:addSpecificEvents which allows zones to define any events they want to spawn in case the spawnZoneSpecificEvents check is met. Specifically, I use this to make the base game gremlin events show up in my zone.

And finally, the big deal:
Adds a new Zone, the gremlin Camp which has four main changes. 3 new encounters, 2 elites, specific event spawns, replacing shops with events and disabling the sleep option at rest sites.

Since I didn't want to create new monsters, I mostly just use the atPreBattle hook to modify the monsters based on the current encounter.


src/main/java/spireMapOverhaul/zones/gremlincamp/GremlinCamp.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private AbstractMonster getGremlin(float x, float y, String prev) {
+        ArrayList<String> gremlinPool = new ArrayList<>();
+        gremlinPool.add(GremlinWarrior.ID);
+        gremlinPool.add(GremlinThief.ID);
+        gremlinPool.add(GremlinFat.ID);
+        gremlinPool.add(GremlinBodyguard.ID);
+        gremlinPool.removeIf(id -> id.equals(prev));
+        String id = Wiz.getRandomItem(gremlinPool, AbstractDungeon.miscRng);
```
erasels, line 112: This is how base game does it as well, personally I would've used monsterRng

src/main/java/spireMapOverhaul/zones/gremlincamp/monsters/GremlinCook.java
```diff
@@ -0,0 +1,142 @@
...
+                                m.increaseMaxHp(healAmt, true);
+                                AbstractDungeon.effectList.add(new HealEffect(m.hb.cX, m.hb.cY, healAmt));
+                            }
+                        });
+                    }
+                });
+                break;
+            case 2:
+                useFastAttackAnimation();
+                Wiz.atb(new ApplyPowerAction(AbstractDungeon.player, this, new PlayerPoisonPower(AbstractDungeon.player, poisonAmt), poisonAmt, AbstractGameAction.AttackEffect.POISON));
```
erasels, line 88: Due to triggering at end of turn, the poison will only start taking effect next turn

src/main/java/spireMapOverhaul/zones/gremlincamp/monsters/GremlinCook.java
```diff
@@ -0,0 +1,142 @@
...
+        //First turn always buffs
+        if(firstMove) {
+            firstMove = false;
+            setMoveShortcut(ALL_BUFF, MOVES[ALL_BUFF]);
+            return;
+        }
+
+        if(!secondMove) {
+            secondMove = true;
+            setMoveShortcut(ALL_BULK, MOVES[ALL_BULK]);
```
erasels, line 114: Does nothing if you manage to kill the grembos by turn 2

src/main/java/spireMapOverhaul/zones/gremlincamp/monsters/GremlinCook.java
```diff
@@ -0,0 +1,146 @@
...
+        super.setUpMisc();
+        this.type = EnemyType.NORMAL;
+    }
+
+    @Override
+    public void takeTurn() { //TODO: Add sounds
+        switch (nextMove) {
+            case 0:
+                useShakeAnimation(Settings.ACTION_DUR_MED);
+                Wiz.atb(new AllEnemyApplyPowerAction(this, strAmt, (m) -> new StrengthPower(m, strAmt)));
```
modargo, line 69: The only fight where Gremlin Cook shows up has it as the last monster to act, right?

Just wanted to be sure about that since otherwise this would change enemy damage mid-turn, which the base game avoids by setting up fights so that strength buffing and vulnerable applying enemies always act last.

src/main/java/spireMapOverhaul/zones/gremlincamp/monsters/GremlinCook.java
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                });
+                break;
+            case 2:
+                useFastAttackAnimation();
+                Wiz.atb(new ApplyPowerAction(AbstractDungeon.player, this, new PlayerPoisonPower(AbstractDungeon.player, poisonAmt), poisonAmt, AbstractGameAction.AttackEffect.POISON));
+                break;
+            case 3:
+                DamageInfo info = new DamageInfo(this, moves.get(nextMove).baseDamage, DamageInfo.DamageType.NORMAL);
+                if (info.base > -1) {
```
modargo, line 92: What's the purpose of this check? Don't we know what the base damage for this move is? I'm guessing this was just copied from somewhere and/or that it was part of some generic check that was copied over for a bunch of moves that weren't all attacks.

So I guess that makes sense, still confusing to read though. I'd want to get rid of it.

src/main/java/spireMapOverhaul/zones/gremlincamp/monsters/GremlinCook.java
```diff
@@ -0,0 +1,146 @@
...
+            possibilities.add(ALL_BUFF);
+        } else {
+            //If alone and late in combat, can buff itself again (helps player)
+            if(GameActionManager.turn > 5 && !lastMove(ALL_BUFF)) {
+                possibilities.add(ALL_BUFF);
+            }
+        }
+
+        // Turn 3+ can start poisoning player
+        if(GameActionManager.turn > 2 && !lastMove(POISON)) {
```
modargo, line 134: Given the firstMove and secondMove checks above, it's always turn 3+ if you get to this logic.

Doesn't really matter, but in making sure I understood the logic of these moves I saw the simplification, so I figured I'd comment.

src/main/java/spireMapOverhaul/zones/gremlincamp/monsters/GremlinCook.java
```diff
@@ -0,0 +1,146 @@
...
+            secondMove = true;
+            if(enemies > 1) {
+                setMoveShortcut(ALL_BULK, MOVES[ALL_BULK]);
+                return;
+            }
+        }
+
+        ArrayList<Byte> possibilities = new ArrayList<>();
+        // If not alone, can buff everyone if it wasn't the last move
+        if(enemies > 1 && !lastMove(ALL_BUFF)) {
```
modargo, line 124: This if and else could be combined to `(enemies > 1 || GameActionManager.turn > 5) && !lastMove(ALL_BUFF)`. WIth a comment like "If not alone or alone and late in combat, can buff if it wasn't the last move".

Same outcome, but a bit easier to understand.

src/main/resources/anniv6Resources/localization/eng/GremlinCamp/Powerstrings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "${ModID}:PlayerPoisonPower": {
+    "DESCRIPTIONS": [
+      "At the end of your turn, take #b%s damage, then reduce this by #b1."
```
modargo, line 4: If  you want to be consistent with the base game wording.
```suggestion
      "At the end of your turn, take #b%s damage, then reduce #yPoison by #b1."
```

src/main/java/spireMapOverhaul/zones/gremlincamp/GremlinCamp.java
```diff
@@ -0,0 +1,202 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static spireMapOverhaul.SpireAnniversary6Mod.makeID;
+import static spireMapOverhaul.SpireAnniversary6Mod.zoneEvents;
+
+public class GremlinCamp extends AbstractZone implements EncounterModifyingZone, CombatModifyingZone, CampfireModifyingZone, ModifiedEventRateZone {
```
modargo, line 45: How do you intend for this zone to play out in terms of balance? It looks like it's mostly a balance-neutral zone with various tradeoffs (? nodes instead of shop nodes and no resting, but the fights don't seem that bad). Not making changes to rewards also suggests that it should aim for being balance neutral.

modargo (commented):
Looking good, I made some comments but they were all very minor. Mostly just slight oddities in the code, where as part of working out what the code did I came up with changes that would make it easier to understand.

I didn't think deeply about balance for these encounters (mostly due to time constraints -- lots of stuff to do). Nothing immediately jumped out as unreasonable, so if there's anything we overlooked, hopefully we catch it during playtesting.

src/main/java/spireMapOverhaul/zones/gremlincamp/GremlinCamp.java
```diff
@@ -0,0 +1,202 @@
...
+    public void replaceRooms(Random rng) {
+        //Replace all shops with event rooms
+        for (MapRoomNode node : this.nodes) {
+            if (node.room instanceof ShopRoom) {
+                node.setRoom(new EventRoom());
+            }
+        }
+    }
+
+    //I think the encounters might be a bit too hard for the first battle
```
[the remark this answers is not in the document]
erasels: Haha, I forgot how crazy A17 Gremlin Wizard is

src/main/java/spireMapOverhaul/zones/gremlincamp/GremlinCamp.java
```diff
@@ -0,0 +1,202 @@
...
+        String id = Wiz.getRandomItem(gremlinPool, AbstractDungeon.miscRng);
+        if (GremlinBodyguard.ID.equals(id)) {
+            return new GremlinBodyguard(x, y);
+        } else {
+            return MonsterHelper.getGremlin(id, x, y);
+        }
+    }
+
+    @Override
+    public void atPreBattle() {
```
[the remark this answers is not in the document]
erasels: Same, kind of sad people didn't do more with it. Seeing base game enemies used in new ways is always fun

src/main/java/spireMapOverhaul/zones/gremlincamp/monsters/GremlinCook.java
erasels: Yep, it always acts last

src/main/java/spireMapOverhaul/zones/gremlincamp/GremlinCamp.java
erasels: Yeah, I consider this a theme park zone that just throws some different stuff at you. Don't think that every zone needs to change the rewards meaningfully.

PR #98, opened 2024-01-16
The Golden Curse

tetraminus:
Combats will drop more rewards, but they cost gold to claim. 
Gold rewards will not drop.
restricted to the city.

Still probably needs some tuning like prices and such, but that's pretty easy to do after the fact.


src/main/java/spireMapOverhaul/zones/goldencurse/RewardPatches.java
```diff
@@ -0,0 +1,103 @@
...
+    public static class RewardItemConstructors {
+        public static void Postfix(RewardItem __instance) {
+            if (!(ZonePatches.currentZone() instanceof GoldenCurseZone)) {
+                return;
+            }
+            int goldAmt = GoldenCurseZone.getRewardCost(__instance);
+
+            RewardItemFields.cost.set(__instance, goldAmt);
+            if (goldAmt > 0) {
+                __instance.text = __instance.text + TEXT[0] + goldAmt;
```
erasels, line 59: This is liable to go over the text length limit or be replaced in certain situations.

How about we just render a gold icon and the number next to it maybe in red) in the bottom right corner of the reward?
Here's an example where we do something like this
https://github.com/erasels/RoR-StS-Mod/blob/7184c76181b3be0c573fc800f9037b70fd9f18e2/RoRStS/src/main/java/riskOfSpire/rewards/ExpensiveLinkedReward.java#L59

src/main/java/spireMapOverhaul/zones/goldencurse/RewardPatches.java
```diff
@@ -0,0 +1,103 @@
...
+            method = SpirePatch.CONSTRUCTOR,
+            paramtypez = {AbstractPotion.class}
+
+    )
+    public static class RewardItemConstructors {
+        public static void Postfix(RewardItem __instance) {
+            if (!(ZonePatches.currentZone() instanceof GoldenCurseZone)) {
+                return;
+            }
+            int goldAmt = GoldenCurseZone.getRewardCost(__instance);
```
erasels, line 55: Why do this in the constructor of reward item instead of just doing it in modifyRewards? That seems simpler to me

erasels (changes requested):

src/main/java/spireMapOverhaul/zones/goldencurse/GoldenCurseZone.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    protected boolean canIncludeTreasureRow() {
+        return false;
+    }
+
+    @Override
+    public boolean canSpawn() {
+        return ActUtil.getRealActNum() == 2;
```
erasels, line 106: Maybe act 2 and 3? I like stuff to be as unrestricted as possible

src/main/java/spireMapOverhaul/zones/goldencurse/GoldenCurseZone.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private final Color color = Color.GOLD.cpy();
+
+    public GoldenCurseZone() {
+        super(ID);
+        this.width = 2;
+        this.maxWidth = 3;
+        this.height = 2;
+        this.maxHeight = 3;
+        iconTexture = ImageMaster.MAP_NODE_ELITE;
```
erasels, line 32: Also, maybe use the golden idol relic as image? Would be pretty fitting

modargo (commented):
Cool idea for a zone. It's nice to have a zone that is purely reward modification, and the prices are roughly in line with what I would have thought. Interested to see how this will work out in practice.

src/main/java/spireMapOverhaul/zones/goldencurse/GoldenCurseZone.java
tetraminus: will do, I totally forgot to change it.

src/main/java/spireMapOverhaul/zones/goldencurse/RewardPatches.java
tetraminus: makes sense. ill change it soon.

PR #99, opened 2024-01-16
feat: Adding toggle for enabling/disabling shaders for Storm and The Fog Zone

EricBartusch:
I passed on making this config work on Broken Space Zone because I haven't heard any complaints about it. I did try for a bit, but stopped when I ran into some issues getting it to work that I didn't feel like dealing with. Can revisit later if need be.


modargo:
Looks good, thanks for adding this. Easy enough to extend it to more effects later.

PR #103, opened 2024-01-16
Keymaster zone

Cany0udance:
A "secret" zone that can only appear if the player has all 3 Keys. Realistically is a rare Act 3 zone but can also technically appear in Act 2.

The modifications to this zone are a strict buff to the player. You could make the argument the downside to the zone appears before the zone itself, as the Keys all require sacrifices to acquire, but I am willing to make it a bit more interesting. Maybe I could replace all event nodes in the zone with normal fights and give all normal fights a burning Elite modifier?

Also, you could argue that this zone breaks the "3 unique features" rule as it really only has 2: A basic combat modifier and a basic reward modifier.


src/main/java/spireMapOverhaul/zones/keymaster/KeymasterZone.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    protected boolean canIncludeEarlyRows() {
+        return true;
+    }
+
+    @Override
+    public void atBattleStart() {
+        forAllMonstersLiving(m -> {
```
erasels, line 55: How about we remove the enemy debuff and just give the player 1 focus as well.

That should be enough and fits themtically with the blue key

src/main/java/spireMapOverhaul/zones/keymaster/KeymasterZone.java
```diff
@@ -0,0 +1,90 @@
...
+            atb(new ApplyPowerAction(m, null, new StrengthPower(m, -2)));
+        });
+        atb(new ApplyPowerAction(AbstractDungeon.player, AbstractDungeon.player, new StrengthPower(AbstractDungeon.player, 1), 1));
+        atb(new ApplyPowerAction(AbstractDungeon.player, AbstractDungeon.player, new DexterityPower(AbstractDungeon.player, 1), 1));
+    }
+
+    @Override
+    public float changeCardUpgradeChance(float curChance) {
+        return 1.0f;
```
erasels, line 64: We also have a hook that allows rare cards to be upgraded in case you want to use that (rare cards in card rewards normally cannot be upgraded)

erasels (changes requested):
Zone looks good to me, I like it.

src/main/java/spireMapOverhaul/zones/keymaster/KeymasterZone.java
Cany0udance: I thought about that, but that point of Focus is useless on every base character besides Defect.

Maybe I should make it so that if the player is Defect, Focus is granted, otherwise the enemies just lose Strength?

Cany0udance:
Accidentally committed a change that I made to my Ambushed zone as well

PR #105, opened 2024-01-16
Keymaster zone

Cany0udance:
New PR, I botched the previous one.


src/main/java/spireMapOverhaul/zones/keymaster/KeymasterZone.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public void atBattleStart() {
+        atb(new ApplyPowerAction(AbstractDungeon.player, AbstractDungeon.player, new StrengthPower(AbstractDungeon.player, 1), 1));
+        atb(new ApplyPowerAction(AbstractDungeon.player, AbstractDungeon.player, new DexterityPower(AbstractDungeon.player, 1), 1));
+        atb(new ApplyPowerAction(AbstractDungeon.player, AbstractDungeon.player, new FocusPower(AbstractDungeon.player, 1), 1));
+    }
+
+    @Override
+    public float changeCardUpgradeChance(float curChance) {
+        return 1.0f;
```
erasels, line 62: If you want rare cards to also be upgraded you'd need to add

```
@Override
    public boolean allowUpgradingRareCards() {
        return true;
    }
```

erasels:
Looks good to me otherwise

src/main/java/spireMapOverhaul/zones/keymaster/KeymasterZone.java
```diff
@@ -0,0 +1,88 @@
...
+        for (AbstractCard card : colorlessCards) {
+            if (card.canUpgrade()) {
+                card.upgrade();
+            }
+        }
+    }
+
+    @Override
+    public void postAddIdleMessages(ArrayList<String> idleMessages) {
+        idleMessages.add("See you at the top!");
```
erasels, line 84: This can't be localized like this, not the end of the world, but you could add it to the UI strings

PR #107, opened 2024-01-16
Grass Zone

SevenDayCandle:
Adds the grass zone and its single current combat modifier. Grass appears in combat, and a random grass grows once for each unspent energy at the end of your turn. When you click on a grass, you launch it at enemies. The type of grass that appears at the start of combat is random. The current grasses are:
- Bell Pepper: Apply (Level) Poison to a random enemy 3 times
- Carrot: Deal 4 damage to a random enemy (Level * 2) times
- Leek: Apply 1 Weak to a random enemy (Level) times
- Onion: Apply 1 Vulnerable to a random enemy (Level) times
- Pumpkin: Gain (Level * 7) block
- Radish: Deal 4 damage to a random enemy (Level * 2) times
- Tomato: Gain (Level * 2) Strength

EDIT: Also added a shop and a reward modifier that increases the chances for relics to be Turnip, Ginger, or Happy Flower

Notes:
- Two utility functions were added to Wiz that are used by this zone


src/main/resources/anniv6Resources/localization/eng/Grass/UIstrings.json
```diff
@@ -0,0 +1,12 @@
+{
+  "${ModID}:Grass": {
+    "TEXT": [
+      "Grass",
+      "The Spire has grass. Grass needs Energy to turn into juicy vegetables.",
```
modargo, line 5: I just merged in code that will show the author for each zone. You should add an entry here with the name you'd like to be credited by. You can pull and look at other zones for examples.

Since you have other text here you'll probably need to adjust the code where they're used since inserting an entry here will shift the indexes, sorry about that!

Also, you should add yourself with the same name to ModTheSpire.json.

erasels (approved):
code looks great

PR #108, opened 2024-01-17
Humility Zone

kiooeht:
Includes all Humility changes except:
- Boss changes, since zones don't affect bosses
- Act 4 changes, since no zones in act 4
- Mind Bloom double boss, since the boss changes aren't included
- The Vulnerable fix Humility normally includes has been altered to only work during runs with zones enabled

Gives a random rare relic when entered.

TODO:
- [x] Zone icon


src/main/resources/anniv6Resources/localization/eng/Humility/UIstrings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "${ModID}:Humility": {
+    "TEXT": [
+      "Humility",
+      "All enemies are #rbuffed in new ways. NL Upon entry, obtain a random #grare #grelic [REMOVE_SPACE]."
```
modargo, line 5: You probably saw my message in #modding-technical but posting here as a reminder that I just merged in code that will show the author for each zone and you should add an entry here for that.

modargo, line 5: Very minor, but given how Humility is a serious difficulty increased, I'd suggest adding the word "dangerous" here. It's something we've used in a few other zones to communicate to the player "hey, we're serious, this zone is a real difficulty increase".
```suggestion
      "All enemies are #rbuffed in new and #rdangerous ways. NL Upon entry, obtain a random #grare #grelic [REMOVE_SPACE]."
```

modargo:
This zone definition makes sense, as do the other zone-specific files. Given that the vast bulk of the other code is from Humility with slight modifications to fit the zone framework, I don't think we need to spend a bunch of time looking through it all. Any issues from the porting process shouldn't be too hard to fix.

We took a similar approach to the port I did of monsters from my act mod; Gk skipped over most of them since it didn't seem worth the time. Plus there will be other last minute zones that are in much more need of scrutiny to catch bugs and issues. So as far as I'm concerned this is good to go once you've tested it to your satisfaction and taken care of the minor TODOs.

src/main/kotlin/spireMapOverhaul/zones/humility/actions/ApplyPowerAllMonstersAction.kt
```diff
@@ -0,0 +1,31 @@
+package spireMapOverhaul.zones.humility.actions
+
+import basemod.interfaces.CloneablePowerInterface
+import com.megacrit.cardcrawl.actions.AbstractGameAction
+import com.megacrit.cardcrawl.actions.common.ApplyPowerAction
+import com.megacrit.cardcrawl.core.AbstractCreature
+import com.megacrit.cardcrawl.dungeons.AbstractDungeon
+import com.megacrit.cardcrawl.powers.AbstractPower
+
+class ApplyPowerAllMonstersAction(
```
erasels, line 10: this is in stslib now, right?

kiooeht: This one works differently, it takes a power instance rather than a lambda that makes powers. It's used to instrument Spiker to change from gaining thorns itself to applying thorns to all enemies by replacing the ApplyPowerAction with this action.
https://github.com/erasels/SpireMapOverhaul/blob/993d474cde6a5d2aa4f46e8c98ed9f1b93c430ac/src/main/kotlin/spireMapOverhaul/zones/humility/patches/beyond/SpikerBuffAll.kt#L21-L27

erasels: Haha, what a madlad, alright.

kiooeht:
Zone should be done. It's hard to exhaustively test as it would require testing every fight both in and out of the zone and the fight command doesn't work properly in zones.

PR #142, opened 2024-01-19
(Ambushed) Reduced zone size, updated UI strings, added Courier support

Cany0udance:
The potential for 4 straight floors of card draw cards could feel bad for players in Act 1 because only the utility of the deck improves when they need block/damage, so reducing height by 25% should help with that frustration.

Also I simply don't have enough combats for the zone, 4 nodes made it very likely for repeats to occur

### DIFF
--- a/src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
+++ b/src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
@@ -58,7 +58,7 @@ public class AmbushedZone extends AbstractZone implements CombatModifyingZone, E
     public AmbushedZone() {
         super(ID, Icons.MONSTER, Icons.SHOP);
         this.width = 3;
-        this.height = 4;
+        this.height = 3;
         initCardDrawCardIDs();
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -235,6 +235,43 @@ private boolean isCardForShop(String cardID, AbstractCard.CardRarity rarity, Abs
...
+        if (replacementCard == null) {
+            // You can decide how to handle this - return null, a default card, etc.
+            return null; // or some default behavior
+        }
+
+        return replacementCard;
+    }
+
+    private AbstractCard getCardForCourier(AbstractCard.CardType type, boolean isColored) {
+        HashSet<String> selectedCardIDs = new HashSet<>(); // To avoid selecting the same card
```
erasels, line 253: You never make use of this and it doesn't nothing to prevent selecting the same card.

src/main/java/spireMapOverhaul/zones/ambushed/AmbushedZone.java
```diff
@@ -235,6 +235,43 @@ private boolean isCardForShop(String cardID, AbstractCard.CardRarity rarity, Abs
...
     }
 
+    @Override
+    public AbstractCard getReplacementShopCardForCourier(AbstractCard purchasedCard) {
+        boolean isColored = purchasedCard.color != AbstractCard.CardColor.COLORLESS;
+        AbstractCard replacementCard = getCardForCourier(purchasedCard.type, isColored);
+
+        // Handle the case where no replacement card is found
+        if (replacementCard == null) {
+            // You can decide how to handle this - return null, a default card, etc.
```
erasels, line 245: Please remove the default comments if you don't need them to explain complex beahvior

erasels (changes requested):
Yeah, the height change makes sense

PR #143, opened 2024-01-19
Windy Zone go Woosh

iryan0702:
Windy zone:

+ Combat modifier: Enemies are defeated when below 10% hp, reducing the gold reward amount equal to their hp.
+ Event – Windy Bridge: The player may grab a hidden card/see its cost/rarity/type at the cost of potentially losing a card/relic/potion/gold, or leave
+ Vfx and Sfx: Dust particles, enemies flying away, and ambient wind noise


src/main/java/spireMapOverhaul/zones/windy/events/WindyBridgeEvent.java
```diff
@@ -0,0 +1,294 @@
...
+    public AbstractRelic getLosableRelic(){
+        ArrayList<AbstractRelic> relics = new ArrayList<>(AbstractDungeon.player.relics);
+        Collections.shuffle(relics, new Random(AbstractDungeon.miscRng.randomLong()));
+        return relics.size() > 0 ? relics.get(0) : null;
+    }
+
+    //identify a random highest tier card that can be lost
+    public AbstractCard getLosableCard(){
+        CardGroup group = CardGroup.getGroupWithoutBottledCards(AbstractDungeon.player.masterDeck);
+        couldLoseCard = group.getRandomCard(AbstractDungeon.eventRng, AbstractCard.CardRarity.RARE);
```
erasels, line 260: Man, events are annoying with the rngs they use. Some use cardRng for card stuff like gremlin match while others use miscRng for basically everything.

Since you went the misc rng route, maybe keep that that way here? eventRng should probably only be used for spawning events

src/main/java/spireMapOverhaul/zones/windy/patches/GoldRewardReductionPatch.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class GoldRewardReductionPatch {
+    public static int combatGoldReduction = 0;
+
+    //on the first call to addGoldToRewards, reduce gold reward param
+    @SpirePatch2(clz = AbstractRoom.class, method = "addGoldToRewards")
+    public static class GoldReductionPatch {
+        @SpirePrefixPatch
+        public static void patch(AbstractRoom __instance, @ByRef int[] gold) {
+            gold[0] = Math.max(0, gold[0]-combatGoldReduction);
```
erasels, line 17: Hmm, since the gold reduction is based on the remaining health of the enemy which is in turn based on how much max health they have, this could result in pretty regular 0 gold rewards in act 3.

What do you think of either removing the reward in that case or capping it to at least 1 gold?

erasels (changes requested):
Thanks for all the comments, this was a pretty easy to understand zone because of them ❤️

src/main/java/spireMapOverhaul/zones/windy/events/WindyBridgeEvent.java
iryan0702: Whoops, missed these ones. I'll set them to miscRng.

src/main/java/spireMapOverhaul/zones/windy/patches/GoldRewardReductionPatch.java
iryan0702: I think removing the reward if it hits 0 gold seems like it'd be the most intuitive without needing extra explanation. A return statement if it hits 0 should work here.

iryan0702:
Committed suggested changes. also it's 7 am i go sleep

PR #169, opened 2024-01-22
Heavenly Clouds zone: change Flighty to remove itself after being triggered and to use a tooltip instead of adding to the card description

modargo:
This is the same way Electric (from the Storm zone) does it, and I think the UX is better. You don't get cards with super tiny text, and it's clear when the mod has been triggered already this combat.


erasels:
Yeah, I think this is the way, thanks for doing that.

PR #171, opened 2024-01-24
Add Smith's Folly

Cany0udance:
Lots of stolen code here.

The main gimmick is that all monsters (including Elites) have Burning Elite modifiers. Upon entering and exiting the zone, the rarest card in your deck is Upgraded. Also includes a fairly powerful event.


erasels:
Aw, why did you deete the PR? I just needed a break from working actively on these mods so I haven't gotten around to looking at the code yet

PR #182, opened 2024-01-28
Russian localization

Kelzon212:
Russian translation
I've never used this site, so maybe I did something wrong.


erasels:
Sorry, had to revert the this. The encoding was messed up and the translation couldn't run. Rucodeby, a russian modder is working on a translation and will make a PR when that's done.

PR #193, opened 2024-02-03
fix map gen thingy

Alchyr:
noticed too many paths that quickly reconnect, probably caused by this not properly applying

can and will break in-progress runs though, due to changing map gen


erasels:
Can you give an example? I'm not sure what you mean, personally I wouldn't want to make the zones too exclusive in pathing

Alchyr:
normal map gen avoids this kinda thing 
image
split -> immediate reconnect
code was intended to do the same, but was not applying properly because it just Kept Going when it should not have

erasels:
If this results in more noodle hallways I might revert it but thanks

PR #211, opened 2024-03-10
Add Smith's Folly

Cany0udance:
Previously made a PR for this but it got erased when I nuked the branch that it was on.


erasels:
Oh nice, glad to see it back. i'll get around to looking over this, I promise.

src/main/java/spireMapOverhaul/zones/smithsFolly/SmithsFolly.java
```diff
@@ -0,0 +1,179 @@
...
+        return null;
+    }
+
+    private void upgradeRarestCard() {
+        AbstractPlayer player = AbstractDungeon.player;
+        if (player == null) return;
+
+        // Define the rarity order
+        List<AbstractCard.CardRarity> rarityOrder = Arrays.asList(
+                AbstractCard.CardRarity.SPECIAL,
```
erasels, line 138: I understand the logic but it'd be more intuitive if special were above basic. People expect rare cards to be the rarest ones

src/main/java/spireMapOverhaul/zones/smithsFolly/SmithsFolly.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+
+    @Override
+    protected boolean canIncludeEarlyRows() {
+        return false;
+    }
+
+    @Override
+    public float zoneSpecificEventRate() {
```
erasels, line 175: Could also add the cursed forge (or whatever it's name is) from base game if you wanted to.
```
@Override
    public Set<String> addSpecificEvents() {
        Set<String> baseGameEvents = new HashSet<>();
        baseGameEvents.add(TheEvent.ID);
        return baseGameEvents;
    }
```

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
```diff
@@ -0,0 +1,162 @@
...
+        switch (screen) {
+            case INTRO:
+                switch (buttonPressed) {
+                    case 0:
+                        handleSearingBlowOption();
+                        break;
+                    case 1: // Gain potion slot and one Blessing of the Forge potion
+                        // Increase potion slots
+                        AbstractDungeon.player.potionSlots += 1;
+                        AbstractDungeon.player.potions.add(new PotionSlot(AbstractDungeon.player.potionSlots - 1));
```
erasels, line 68: Does this work with save and reload? Just to be sure

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
```diff
@@ -0,0 +1,162 @@
...
+            searingBlow.upgrade();
+            searingBlow.upgrade();
+            searingBlow.upgrade();
+            AbstractDungeon.effectList.add(new ShowCardAndObtainEffect(searingBlow, (float)Settings.WIDTH / 2.0F, (float)Settings.HEIGHT / 2.0F));
+            this.imageEventText.updateBodyText(DESCRIPTIONS[5]);
+        }
+    }
+
+    private void upgradeSearingBlow(int times) {
+        for (AbstractCard card : AbstractDungeon.player.masterDeck.group) {
```
erasels, line 128: This would upgrade multiple searing blows if you have more than one

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
```diff
@@ -0,0 +1,162 @@
...
+                }
+                // Show visual effects for each upgrade
+                AbstractDungeon.topLevelEffects.add(new ShowCardBrieflyEffect(card.makeStatEquivalentCopy()));
+                AbstractDungeon.topLevelEffects.add(new UpgradeShineEffect((float)Settings.WIDTH / 2.0F, (float)Settings.HEIGHT / 2.0F));
+            }
+        }
+    }
+
+
+    private void replaceStrikesWithSearingBlow() {
```
erasels, line 141: This might be too crazy. Removing two strikes and giving one searing blow+4 or removing all strikes and replacing them with unupgraded searing blows is probably better.

erasels (changes requested):

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
Cany0udance: It does.

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
Cany0udance: That is correct. I don't think this is an issue. Let's say you start Act 1 and Smith's Folly appears. You can *choose* to take every Searing Blow you see (which won't be many because it's an uncommon) for the sweet upgrades, but there's only a CHANCE of the event showing up, and if it doesn't, you're stuck hauling around a bunch of Searing Blows that are about as bad as strikes.

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
Cany0udance: Keep in mind this outcome can only occur in Act 3. Most players will probably have removed most of their Strikes by then, and even if they kept all the Strikes, I don't think a few Searing Blow+4 cards will drastically shift the run (2 cost, 34 damage. Carnage+ for comparison is 2 cost, 28 damage, not too far off). It can MAYBE be a desperate last minute damage solution if you don't have one already or it's just a good pick if you have Snecko Eye but I truly don't think it's that overpowered.

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
erasels: Hm, fair point. I guess it could make for a fun decision

src/main/java/spireMapOverhaul/zones/smithsFolly/events/RansackedForge.java
erasels: I can see your point, but I still think people tend to have 2-3 Strikes in their deck by the end of the run on average. That's still replacing the worst attack in the game with cards that are better than an upgraded rare on Ironclad. That's a very strong upside, I fear to think what degeneracy Silent or Watcher could get up to with it. 

Maybe a compromise, upgrade them twice, that way it's 21 damage, still crazy strong but at least not win-con strong.

Cany0udance: I think 2 cost 21 damage cards towards the end of the run is pretty bad, even if they replace Strikes. Would you settle for Searing Blow+3 instead?

erasels: It's not great but it's a big damage attack, characters that aren't balanced around having that would salivate. You won#t find a higher damage attack on watcher, it's probably the best attack on her and Silent still.

And it's certainly better than a strike would ever be.

Cany0udance: Ah, I didn't even think about what you could do with Watcher's stances. I block that character out of my mind.

Alright, I decided to make all the options +2 rather than +4 so they're uniform. That shouldn't be too powerful.

erasels:
Something was messed up with the branch and updating it from main removed a bunch of content. I manually readded the zone with [8d383f52bfd7ca3b4a5656d234341233c46e8e06 ](https://github.com/erasels/SpireMapOverhaul/commit/8d383f52bfd7ca3b4a5656d234341233c46e8e06), I think that's everything? I'd appreciate if you could fly over it as well.

Cany0udance:
Looks good to me

PR #212, opened 2024-03-10
Adjustable Biome Frequency

Cany0udance:
Ignore the scuffed commit history.

This introduces a setting in the mod config menu which allows players to choose how many biomes they want to appear per Act. Options are 0-1, 1-2, 2-3, 3-4, and 4-5.


src/main/java/spireMapOverhaul/BetterMapGenerator.java
```diff
@@ -78,18 +78,18 @@ public ArrayList<ArrayList<MapRoomNode>> generate(Random rng, int width, int hei
                 }
             }
 
-            outer:
-            while (rng.randomBoolean(zoneRate) && activeZones.size() < pathDensity) {
-                //zoneRate *= ZONE_FALLOFF_RATE;
-                while (!possibleZones.isEmpty()) {
-                    zone = possibleZones.remove(rng.random(possibleZones.size() - 1)).copy();
+            outer: while (rng.randomBoolean(zoneRate) && activeZones.size() < pathDensity) {
```
erasels, line 81: Think  prefer a newline after outer

erasels (changes requested):
Could you also please add the english string to the other localizations so they don't crash

PR #214, opened 2024-03-16
Candyland Background + additional framework

Cany0udance:

PR #215, opened 2024-03-16
Revert "Candyland Background + additional framework"

erasels:
Reverts erasels/SpireMapOverhaul#214

Alright, something went very wrong with that update from main


erasels:
Alright, I'm messing up even more 😆

PR #217, opened 2024-03-17
Smiths's Folly rus 

Kelzon212:
Translation into Rus and string in UIstring


src/main/resources/anniv6Resources/localization/rus/UIstrings.json
```diff
@@ -7,7 +7,8 @@
       "Включено",
       "Делает иконки биомов больше но срезает форму биомов.",
       "Не повторять биомы в ходе забега.",
-      "Включить шейдеры"
+      "Включить шейдеры",
+      "Biomes per Act"
```
erasels, line 11: Please feel free to translate Biomes per Act as well, this was just the english string so it doesn't crash

erasels (changes requested):
Thank you for the trnaslation!

PR #226, opened 2024-03-29
Korean localization

moviejw:
Korean localization for this mod.
I'm not familiar with modding, so I'm not sure how to handle the keywords and prefixes in tooltips in the in-game. So I think there might be cases where it doesn't show up properly. I'd appreciate feedback on that or an environment where I can test it myself.


erasels:
For keywords, they need to be prefixed with the modid and have one of the names defined in the names array.
here an example:
```
[
  {
    "ID": "${ModID}:Verdict",
    "PROPER_NAME": "평결",
    "NAMES": ["평결", "평결보다"],
    "DESCRIPTION": "#y평결을 가진 대상은 대상의 턴 시작 시 현재 체력이 #y평결보다 작다면 사망합니다. 아니라면, #y평결을 #b5 얻습니다."
  }
]
```
So if you wanted to use Verdict, you'd use this in card text
`"DESCRIPTION": "Blah blah ${ModID}:평결 blah.`

moviejw:
I would like to ask you a few more questions.

I don't think I fully understood your description, so I'll make an example. In the case of Candylands/Cardstring.json, should I change the fleeting in the card description to ${ModID}:fleeting, ${ModID}:일회용, or just 일회용? 

And ${ModID}: is it for all situations where keywords are applied? Some biomes contain '${ModID}:' in id:, and others don't.

Also, Keywordstring.json contains the NAMES array, and I thought it would work to write down the same keywords so that they can be referenced in situations where they are used in different forms in a sentence. Should I just write all the forms that are used in Cardstring.json and so on? 

Last question, why does metallicize or buffer that exist in vanilla games exist as keywords?

moviejw:
I haven't received an answer yet, so I've revised the rest of it as I understand it so far. I'd like to see it applied to ingame.
Additionally, I'm not used to github, so I think I made a mistake during uploading maybe.
(programmer's world is too complicated)
If there's anything I need to fix, I'll try again

erasels:
> should I change the fleeting in the card description to ${ModID}:fleeting, ${ModID}:일회용, or just 일회용?
I believe fleeting is a keyword from StSLib so you'd want to use 일회성 as defined here https://github.com/kiooeht/StSLib/blob/master/src/main/resources/localization/stslib/kor/legacyKeywords.json
For keywords from this mod, you'd want to always use ${ModID}: + one of the defined names in the korean keyword strings file the keyword is in.

${ModID}: will be replaced by the modid when compiled, so after compilation it'll be anniv6: this is used as a prefix for most things to prevent overlap with other mods. (For example, if two mods had a keyword called Fleeting, then the one loaded later would overwrite the keyword from the earlier one.) It doesn't have to be used for Biome IDs since biomes are something original to this mod and won't be replicated by other mods, some people just prefer to have them though. 

> Should I just write all the forms that are used in Cardstring.json and so on?
Pretty much, everything in the names list will be recognized as belonging to that keyword when used in the cardtext (in combination with the ${ModID}: prefix)

> \u00a0'
I think that's a non breaking space, should be fine to leave.
> [REMOVE_SPACE]
This is used in certain strings (basically everything that's not a card string) for formatting, if you use #y to highlight a word yellow, it will highlight everything until the next space so `#yExhaust, unless xyz` the , after Exhaust would be yellow. To prevent that people use `Exhaust [REMOVE_SPACE], unless xyz` the [REMOVE_SPACE] will be removed before the user can see it and the , won't be yellow since there was a space even if it got removed later.

> why does metallicize or buffer that exist in vanilla games exist as keywords?
Because base game didn't create keywords for them. For metallicize, the card just explains what the power does, it doesn't say Gain 5 Metallicize. To not bloat text modders like making keywords for them.

> I haven't received an answer yet, so I've revised the rest of it as I understand it so far. I'd like to see it applied to ingame.
My bad, was on easter holiday.

moviejw:
Thank you for your response.
I've updated it again.

erasels:
Alrighty thanks, do you want to be credited as moviejw or would you prefer a different name?

moviejw:
I didn't know the comments were coming up even after the request was merged, so I just checked your comment.
I want to be credit as moviejw.

PR #241, opened 2024-04-18
Improved Gravewood Grove visuals(?)

Cany0udance:
Biome now is properly desaturated

Edit: Maybe this isn't actually better, the brown tint from before does have some charm to it. Either way, you decide, feel free to reject this


Cany0udance:
Old vs new

image
image

erasels:
Oh boy, can't wait for people to complain about a black screen due to a shader like with Wildfire! Looks great though, thanks

PR #266, opened 2025-02-08
Changed rendering logic for zones on the map

Cany0udance:
Previously, for me at least, a new issue emerged where some biomes would have their transparencies heavily increased. See below, there are three zones: Gremlin Camp, where the zone outline is entirely invisible, Heavenly Clouds, where the zone outline is BARELY visible, and Mana Surge, where it looks fine.

image

This change makes it so that all biomes have consistent transparency:

image

Again, this was a personal issue and I'm not sure how many others have experienced the same transparency discrepancies, but I hope there's some others out there who are helped by this.

### DIFF
--- a/src/main/java/spireMapOverhaul/abstracts/AbstractZone.java
+++ b/src/main/java/spireMapOverhaul/abstracts/AbstractZone.java
@@ -77,7 +77,7 @@ public abstract class AbstractZone {
     protected float labelX = 0, labelY = 0;
 
     public FrameBuffer zoneFb = null;
-    private TextureRegion shapeRegion = null;
+    public TextureRegion shapeRegion = null;
     public Texture iconTexture;
 
     public List<Hitbox> hitboxes;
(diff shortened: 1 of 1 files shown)


src/main/java/spireMapOverhaul/abstracts/AbstractZone.java
```diff
@@ -211,8 +211,15 @@ public void renderOnMap(SpriteBatch sb, float alpha) {
             }
             float anchorX = x * SPACING_X + OFFSET_X - ZoneShapeMaker.FB_OFFSET;
             float anchorY = y * Settings.MAP_DST_Y + OFFSET_Y + DungeonMapScreen.offsetY - ZoneShapeMaker.FB_OFFSET;
-            sb.setColor(getAdjustedColor().cpy().mul(1f, 1f, 1f, alpha*0.5f));
+
+            Color oldColor = sb.getColor().cpy();
```
erasels, line 215:
```suggestion
            Color oldColor = sb.getColor()
```
No need to make a copy if we're just going to slot it back in after

src/main/java/spireMapOverhaul/abstracts/AbstractZone.java
```diff
@@ -211,8 +211,15 @@ public void renderOnMap(SpriteBatch sb, float alpha) {
...
+
+            Color zoneColor = getAdjustedColor().cpy().mul(1f, 1f, 1f, alpha * 0.5f);
```
erasels, line 217: I'll be honest, I don't think this does anything, you probably just fixed the issue by saving and reloading. However, changing the color back is good practice, so there's value in this change regardless.

erasels (changes requested):

src/main/java/spireMapOverhaul/abstracts/AbstractZone.java
Cany0udance: > I'll be honest, I don't think this does anything, you probably just fixed the issue by saving and reloading. 
What do you mean by this? I spent hours playing Spire Biomes recently with and without this change to confirm it works. It's not just a simple "save and quit to refresh" issue.

PR #267, opened 2025-02-08
Fixed a bug where the zone outlines would disappear between Acts if Minty Spire's minimap was active

Cany0udance:
Note that the only quirk I could find with this is that between Acts, if the minimap is active, it takes a short amount of time for the zone outlines to kick in (2 or so frames from the FRAME_DELAY), but that's definitely an improvement from them not rendering at all.


modargo:
Thanks for contributing this fix!

PR #274, opened 2025-05-14
Cafe merchants

antemercado:
Some of our favorite merchants from the Spire Cafe have set up shop outside the cafe and in the rest of the spire!
image
image

This was a surprisingly easy task. So much tot he point that I'm suspicious I'm missing something. If someone who is familiar with both projects could review that would be great.

This PR relies on some util functions implemented in a PR for Spire Cafe. This should not be merged and deployed until those functions are first reviewed. The code could be refactored to handle it without those helper functions if needed.

https://github.com/erasels/SpireCafe/pull/139


modargo (commented):
This is a fun idea and it's neat to see another single-node zone. There's a lot of variance in how good the Cafe merchants are and you can't rely on getting a specific merchant, so this should be fine in terms of balance. It will be a gamble going into it.

The code looks pretty good overall. I've left a few comments, mainly about a few potential issues to guard against.

